PR #2, opened 2016-05-27
google and numpy docstring support

daviskirk:
Hi,
I needed to add support for the google docstring format for a different project.  
Since [napoleon](http://sphinxcontrib-napoleon.readthedocs.io/en/latest/), the sphinx extension which enabled you to use google docstrings in the first place also supports numpy docstrings I thought I might add them as well. 
Let me know if you are interested in this, I'm also fine with maintaining this seperately for myself. 

In any case, thanks alot for making this in the first place!

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -33,6 +33,10 @@ to this:
         return '{} {}'.format(value, unit)
 
 
+There is also support for google docstrings or numpy docstrings with help of the napoleon
+`napoleon sphinx extention http://sphinxcontrib-napoleon.readthedocs.io/en/latest/`_.
+
+
 Installation and setup
 ----------------------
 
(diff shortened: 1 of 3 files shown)


agronholm:
Sorry to say but I'm not interested in any docstring support or other workarounds. My focus is purely on Python 3 and type annotations. The only acceptable exception to this would be the PEP 484 type comments for backwards compatibility, but I honestly don't know how to implement those.

I would really like this to be integrated in Sphinx itself so I wouldn't have to maintain it anymore, but seeing as how it seems impossible to get even simple bug fixes merged into it, I don't have high hopes for that.

daviskirk:
No problem, I understand completely.

PR #6, opened 2016-09-27
Removed return type of __init__ and fixed some generic type parameters missing.

wheerd:
Some generic types like `Mapping[str, int]` only have `__args__` but not `__parameters__`, so they didn't get the type arguments. Fixed that, but the generic arguments probably need a more general overhaul (when mixing generic and non-generic type parameters, this probably still doesn't work (e.g. for `Mapping[T, int]`).

Also, mypy forces you to have your `__init__` annotated with `-> None`, which was getting added to the docs. Removed that, because it is unnecessary and looks ugly.

### DIFF
--- a/sphinx_autodoc_typehints.py
+++ b/sphinx_autodoc_typehints.py
@@ -78,7 +78,7 @@ def process_docstring(app, what, name, obj, options, lines):
 
             if argname == 'return':
                 if what in ('class', 'exception'):
-                    continue # Don't add return type None from __init__
+                    continue  # Don't add return type None from __init__
                 insert_index = len(lines)
                 for i, line in enumerate(lines):
                     if line.startswith(':rtype:'):
--- a/sphinx_autodoc_typehints.py
+++ b/sphinx_autodoc_typehints.py
@@ -30,9 +30,14 @@ def format_annotation(annotation):
                 ellipsis = annotation.__tuple_use_ellipsis__
             else:
                 params = getattr(annotation, '__parameters__', None)
+                if not params:
+                    params = getattr(annotation, '__args__', None)
+                if params is Ellipsis:
+                    ellipsis = True
+                    params = []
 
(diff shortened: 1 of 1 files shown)


agronholm:
Do you know how I could write a test suite for this? I'm proficient with py.test but I'm not sure how to properly test Sphinx extensions. I'm inclined towards merging your PR but manual testing is a drag.

wheerd:
You should be able to test your hooks just like normal functions, i.e. provide parameters are validate the result. I can recommend to use [ddt](https://github.com/txels/ddt) to easily create a lot of those tests.

Then to mock objects from autodoc,/Sphinx (which I think you are not really using anyway) you can use [unittest.mock](https://docs.python.org/3/library/unittest.mock.html). I am usually using the builtin unittest package, but I think you can use both with py.test as well.

Here is an example of how you could test `process_docstring()`:
```python
import unittest
from unittest.mock import patch
from sphinx.ext.autodoc import Options
from ddt import ddt, unpack, data

from sphinx_autodoc_typehints import process_docstring

def f1(a: int, b: str) -> int: ...

@ddt
class ProcessDocstringTest(unittest.TestCase):
    @unpack
    @data(
        ('function', f1, [], [':rtype: :class:`int`']),
        ('function', f1, [':param a: Fubar'], [':type a: :class:`int`', ':param a: Fubar', ':rtype: :class:`int`']),
    )
    def test_find_cycle(self, what, obj, lines, result):
        with patch('sphinx.application.Sphinx') as Application:
            app = Application.return_value
            process_docstring(app, what, obj.__name__, obj, Options(), lines)
            self.assertListEqual(result, lines)

if __name__ == '__main__':
    unittest.main()
```

agronholm:
I've incorporated the `__init__` change and credited you in the changelog. Merging the patch directly was problematic. Thank you.

PR #7, opened 2016-10-05
Reworked format_annotation()

wheerd:
I reworked `format_annotation()` to work better with Generic types and correctly generate links for the type arguments. Some examples, that didn't work before:

``` python
Optional[Union[int, str]]
Callable[..., int]
Callable[[str, str], int]
Mapping[T, int]
Mapping[str, str]
```

Also string type hints now work correctly:

``` python
class A:
  def x(self) -> 'A':
    return self
```

This will generate a link to the A class now.
I can provide some tests for `format_annotation()` if you like.

### DIFF
--- a/sphinx_autodoc_typehints.py
+++ b/sphinx_autodoc_typehints.py
@@ -18,19 +18,17 @@ except ImportError:
 
 def format_annotation(annotation, obj=None):
     if isinstance(annotation, type):
-        qname = sys.version_info[:2] >= (3, 3) and annotation.__qualname__ or annotation.__name__
-
         # builtin types don't need to be qualified with a module name
         if annotation.__module__ == 'builtins':
-            if qname == 'NoneType':
+            if annotation.__qualname__ == 'NoneType':
                 return '``None``'
             else:
-                return ':class:`%s`' % qname
+                return ':class:`%s`' % annotation.__qualname__
 
         params = None
 
         # Check first if we have an TypingMeta instance, because when mixing in another meta class,
-        # some informatiion might get lost.
+        # some information might get lost.
         # For example, a class inheriting from both tuple and Enum ends up not having the
         # TypingMeta metaclass and hence none of the Tuple typing information.
         if isinstance(annotation, TypingMeta):
@@ -66,7 +64,7 @@ def format_annotation(annotation, obj=None):
                     # Union[T,None]   => Optional[T]
                     # Union[T,U,None] => Optional[Union[T, U]]
                     if type(None) in (annotation.__union_set_params__ or tuple()):
-                        qname = 'Optional'
+                        annotation.__qualname__ = 'Optional'
                         params.remove(type(None))
                         if len(params) > 1:
                             params = [Union[tuple(params)]]
@@ -110,7 +108,7 @@ def format_annotation(annotation, obj=None):
                 return format_annotation(annotation.__type__, obj)
 
         generic = params and '\\[%s]' % ', '.join(format_annotation(p, obj) for p in params) or ''
-        return ':class:`~%s.%s`%s' % (annotation.__module__, qname, generic)
+        return ':class:`~%s.%s`%s' % (annotation.__module__, annotation.__qualname__, generic)
 
     # _TypeAlias is an internal class used for the Pattern/Match types
     # It represents an alias for another type, e.g. Pattern is an alias for any string type
--- a/sphinx_autodoc_typehints.py
+++ b/sphinx_autodoc_typehints.py
@@ -26,7 +26,6 @@ def format_annotation(annotation, obj=None):
                 return ':class:`%s`' % annotation.__qualname__
 
         params = None
-
         # Check first if we have an TypingMeta instance, because when mixing in another meta class,
         # some information might get lost.
         # For example, a class inheriting from both tuple and Enum ends up not having the
@@ -35,16 +34,13 @@ def format_annotation(annotation, obj=None):
             # Since Any is a superclass of everything, make sure it gets handled normally.
             if annotation is Any:
                 pass
-
             # Generic classes have type arguments
             elif isinstance(annotation, GenericMeta):
                 params = annotation.__args__
-
                 # Make sure to format Generic[T, U, ...] correctly, because it only
                 # has parameters but nor argument values for them
                 if not params and issubclass(annotation, Generic):
                     params = annotation.__parameters__
-
             # Tuples are not Generics, so handle their type parameters separately.
             elif issubclass(annotation, Tuple):
                 if annotation.__tuple_params__:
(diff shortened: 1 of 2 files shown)


sphinx_autodoc_typehints.py
```diff
@@ -1,40 +1,120 @@
...
 
+try:
+    from typing import ClassVar
+except ImportError:
+    ClassVar = None
+
+    
+def format_annotation(annotation, obj=None):
+    if isinstance(annotation, type):
+        qname = sys.version_info[:2] >= (3, 3) and annotation.__qualname__ or annotation.__name__
```
agronholm: Why this dance with `__qualname__`? Python 3.3 is the minimum supported version anyway (as evident from the classifiers in setup.py).

sphinx_autodoc_typehints.py
```diff
@@ -1,40 +1,120 @@
...
-                    annotation = Optional
-                    params = (params[0],)
-            elif annotation.__qualname__ == 'Tuple':
-                params = annotation.__tuple_params__
+                return ':class:`%s`' % qname
+
+        params = None
+
+        # Check first if we have an TypingMeta instance, because when mixing in another meta class,
+        # some informatiion might get lost.
```
agronholm: typo

sphinx_autodoc_typehints.py
```diff
@@ -1,40 +1,126 @@
...
 
+try:
+    from typing import ClassVar
+except ImportError:
+    ClassVar = None
+
+
+def format_annotation(annotation, obj=None):
+    if isinstance(annotation, type):
+        qname = sys.version_info[:2] >= (3, 3) and annotation.__qualname__ or annotation.__name__
```
agronholm: Why this dance with `__qualname__`? The project only supports Python 3.3+ anyway, as is evident from the classifiers in setup.py.

sphinx_autodoc_typehints.py
```diff
@@ -1,40 +1,126 @@
...
+        # some informatiion might get lost.
+        # For example, a class inheriting from both tuple and Enum ends up not having the
+        # TypingMeta metaclass and hence none of the Tuple typing information.
+        if isinstance(annotation, TypingMeta):
+            # Since Any is a superclass of everything, make sure it gets handled normally.
+            if annotation is Any:
+                pass
+
+            # Generic classes have type arguments
+            elif isinstance(annotation, GenericMeta):
```
agronholm: No gaps between if...elif...else blocks!

agronholm:
Also, what do you expect `Optional[Union[int, str]]` to render as? It should render as `Union[int, str, None]`.

wheerd:
I expect it to render as `Optional[Union[int, str]]`. Unions are flattened, so if I write `Union[Optional[int], str]` or `Optional[Union[int, str]]` it will become `Union[int, str, None]` (or `Union[int, None, str]` in the first case) . I would expect to always get an `Optional`, when `None` occurs in the `Union`.

agronholm:
The tests look promising. I will take a better look at all this on the coming weekend.

agronholm:
Or maybe the next weekend. Been very busy :/

agronholm:
OK, I've had time to look at this. Here are my thoughts so far:

* I still don't like `Optional[Union[...]]`. Unless you can find quotes from higher ups that say this is how it should look like, I'm going to exercise my administrative powers to overrule this.
* I'm having doubts about `_ForwardRef` and `_TypeAlias`. I don't like the fact that the code imports classes meant for internal use as the API can break any time. Won't `get_type_hints()` at least eliminate the forward references?
* The tests only test `format_annotation`, not `process_signature` or `process_docstring`. Those are the functions I would like to see tested (and is the reason why I hadn't written any tests myself yet).

Smaller housekeeping:

* The call to `test_format_annotation_with_obj()` at the bottom should be removed
* Use PEP 8 grouping for imports in the test module

Thank you for your efforts so far. This should be a notable improvement to the functionality once merged!

wheerd:
> I still don't like `Optional[Union[...]]`. Unless you can find quotes from higher ups that say this is how it should look like, I'm going to exercise my administrative powers to overrule this.

I am only trying to be as close to how you would write it in the code. There I would prefer `Optional[Union[int, str]]` over `Union[int, str, None]`.

> I'm having doubts about `_ForwardRef` and `_TypeAlias`. I don't like the fact that the code imports classes meant for internal use as the API can break any time. Won't get_type_hints() at least eliminate the forward references?

True. The first one is automatically resolved, if you use `get_type_hints()`. However, for the formatted base list in #8, you still only get strings. Unfortunately, this functinality is not exposed anywhere. Any recursive class-based data structure will need to use string references in a base class:

```python
T = typing.TypeVar('T')
U = typing.TypeVar('U')
class DictTree(typing.Dict[T, typing.Union[U, 'DictTree[T, U]']], typing.Generic[T, U]):
    pass
```

The second one can be covered by explicitly checking every instance of `_TypeAlias` which for now are `Pattern` and `Match`. I thought that in case there are more of this kind added later, they are already covered this way. But maybe it is better to avoid using the internals there.

Alas, the code already heavily relies on internals anyway (`__tuple_params__` etc.). Since none of the classes in the `typing` module expose the parameters, the code will break if they change the internal structure. So I am unsure whether this is a problem.

> The tests only test `format_annotation`, not `process_signature` or `process_docstring`. Those are the functions I would like to see tested (and is the reason why I hadn't written any tests myself yet).

Well those functions are not changed by this pull request,, so I don't think that tests for them should be part of it either.

I hope I have taken care of the housekeeping stuff with my latest commit.

agronholm:
Have you checked whether this code works with Python 3.6? It wreaks havoc with my other project (typeguard).

wheerd:
No I haven't.

agronholm:
I'm pretty sure this doesn't work with Python 3.6. I released typeguard 2.1.0 a short while ago which underwent major refactoring to make it work with the totally revamped typing module.

wheerd:
It doesn't work with python 3.6. I have made a fixed version for `sphinx_autodoc_napoleon_typehints` in f31927c1aed64850b79ce8bbb40856c025fba8f4. However, I still need to improve upon that, because that version does not work for earlier python versions anymore. I am hoping, that they will release a fixed version of the typing module for 3.5, but probably not. The 3.5 module has some serious bugs that make it hard to use. Since I am using the napoleon extension, I am not more interested in getting the type hint support in there directly.

agronholm:
I will fix it the way I fixed my other project, typeguard.

agronholm:
I've been making some progress with the 3.6 upgrade. I'm using your test suite but it seems unlikely that I'll use your full patch.

wheerd:
Okay, the tests are not completely correct anymore either, as I discovered that most of the typing classes are actually documented with a `data` role instead of `class`. This applies to `Any`, `Union`, `Callable`, `Tuple` and `Optional`. An updated test suite can be found [here](https://github.com/wheerd/sphinx-autodoc-typehints/blob/f31927c1aed64850b79ce8bbb40856c025fba8f4/tests/test_autodoc.py#L40).

agronholm:
I have integrated most of your test suite and credited you in the changelog. I will not pull this PR as I ended up with significantly different code when fixing the problem myself. Thank you for your efforts!

PR #17, opened 2017-07-21
Add CI, and comment support

gaborbernat:
Hello, 

On my fork I've some changes:
- I have added CI (and some integration tests - e.g. simulate on actual Sphinx projects via sphinx.tester) on my fork,
- added support for type hint comments for functions (see PEP-484), and PyCharms documentation on this (this is so that libraries that aim both Python 2/3 support can use this plugin) e.g. https://github.com/gaborbernat/sphinx-autodoc-typehints/blob/master/tests/roots/test-comments/test_code/__init__.py#L18,
- added pylint to test style,
- dropped backport in favor of typing library (works just as well under earlier Pythons),
- also dropped support for Python 3.3.

Not sure if we want to merge this back to the upstream, or just keep it as a fork. Please advise.

Thanks,


agronholm:
There is no way I will accept such broad changes without a very good reason. Try starting from the type comment support – if it's maintainable code, I might accept it in a PR.

agronholm:
Also, when you write your commit messages, stick with the upstream style.

gaborbernat:
@agronholm I've done too many changes and would take significant amount of work to split it up (lately now I've also added to option to alias classes, so that you can expose sub-module at root level e.g.); so I think I'll just fork it under a new name in PyPi.

PR #19, opened 2017-08-10
Add explicit py domain to added directives

emonty:
If a project is using an explicit sphinx default domain, then the inserted :class: and :data: wind up being not found. Since this is an extension specifically for annotating python code, be explicit with the directives adding a :py: domain prefix to them.

### DIFF
--- a/sphinx_autodoc_typehints.py
+++ b/sphinx_autodoc_typehints.py
@@ -35,30 +35,30 @@ def format_annotation(annotation):
         if annotation.__qualname__ == 'NoneType':
             return '``None``'
         else:
-            return ':class:`{}`'.format(annotation.__qualname__)
+            return ':py:class:`{}`'.format(annotation.__qualname__)
 
     annotation_cls = annotation if inspect.isclass(annotation) else type(annotation)
     if annotation_cls.__module__ in ('typing', 'backports.typing'):
         params = None
-        prefix = ':class:'
+        prefix = ':py:class:'
         extra = ''
         class_name = annotation_cls.__qualname__
         if annotation is Any:
-            return ':data:`~typing.Any`'
+            return ':py:data:`~typing.Any`'
         elif annotation is AnyStr:
-            return ':data:`~typing.AnyStr`'
+            return ':py:data:`~typing.AnyStr`'
         elif isinstance(annotation, TypeVar):
             return '\\%r' % annotation
         elif class_name in ('Union', '_Union'):
-            prefix = ':data:'
+            prefix = ':py:data:'
             class_name = 'Union'
             if hasattr(annotation, '__union_params__'):
                 params = annotation.__union_params__
             else:
                 params = annotation.__args__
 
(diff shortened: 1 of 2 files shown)


setup.py
```diff
@@ -1,3 +1,4 @@
+# -*- coding: utf-8 -*-
```
agronholm: What is this doing here?

tox.ini
```diff
@@ -20,3 +20,6 @@ commands = python -m pytest {posargs}
 deps = flake8
 commands = flake8 sphinx_autodoc_typehints.py test_sphinx_autodoc_typehints.py
 skip_install = true
+
+[flake8]
```
agronholm: Any flake8 options go to setup.cfg. Also, this project uses a maximum line width of 99 columns, which is the maximum allowed by PEP 8. This is also obvious from setup.cfg. Therefore ignoring E501 is not acceptable. The E241 error is ignored in the default configuration (which is how flake8 have passed previously).

agronholm:
Apart from my comments, this seems reasonable. When you've addressed the problems I will merge this.

setup.py
agronholm: Ok, I just read the comment for the commit. I'm still -1 on adding this. You could install `tox` on Python 3 instead and do away with this problem that way.

agronholm:
If you have ideas on how to fix the backports tests, I would appreciate it.

setup.py
emonty: Without this the following error occurs when I run tox:

  SyntaxError: Non-ASCII character '\xc3' in file setup.py on line 14, but
  no encoding declared; see http://python.org/dev/peps/pep-0263/ for
  details

Triggered, I believe, by the umlaut in your name.

tox.ini
emonty: Aha! That makes SO much more sense. Proper fix coming.

setup.py
agronholm: Yes, but if you uninstall tox and install it again on Python 3, this problem will go away.

emonty:
I looked at the backports issue this morning for a while and eventually found:

https://bugs.python.org/issue28339

It seems, if I'm reading it right, that there is an issue sphinx.ext.autodoc is hitting in its code that has been fixed in 3.5, 3.6 and 3.7 but not 3.4. I can reproduce it locally in tox, but oddly not in a directly created virtualenv. I'm still poking to see if there is a workaround, but so far I haven't found one.

setup.py
emonty: Ah - so it does. Neat.

agronholm: Ok so drop this one and make the tests pass and I'll merge.

agronholm:
I'll point out that there are unresolved test failures, unrelated to backports. Please fix those and I will merge this then.

agronholm:
Thanks!

emonty:
Thank you! I'm excited to start using this in our project.

Issue #23, opened 2017-11-14
static and class methods in a custom exception lose their first argument


scottbelden:
If I have something like this:

```python
class MyException(Exception):
    @classmethod
    def from_name_and_number(cls,
                             name: str,
                             number: int):
        """
        Args:
            name: name
            number: number
        """
        return cls(f'{name}: {number}')

    @staticmethod
    def is_same(name: str,
                number: int) -> bool:
        """
        Args:
            name: name
            number: number
        """
        return str(number) == name
```

When I generate the documentation I get the following:
```
classmethod from_name_and_number(number)
    Parameters: * name(str) - name
                * number(int) - number

static is_same(number)
    Parameters: * name(str) - name
                * number(int) - number

    Return type: bool
```

Notice that both the functions look like they only take in a `number` argument. I think this is specific to `Excpetions` as I don't think I see the same issues for other methods.

For reference, if I run without this extension I get something like the following:
```
classmethod from_name_and_number(name: str, number: int)
    Parameters: * name - name
                * number - number

static is_same(name: str, number: int) → bool
    Parameters: * name - name
                * number - number
```

scottbelden:
I'm pretty sure it was the case, but I'd have to set up the example I posted above again to know for sure. As it turns out I realized that I didn't really need the static or class methods anyways, so this isn't really important to me anymore, but I didn't close it as I believe it is an issue.

agronholm:
Oddly enough, the problem seems to be easily fixable for class methods, as they appear as bound methods. But I have not yet found a way to distinguish between static and instance methods.

agronholm:
Fixed via 4fd7cff2abe446a10567f75fab2b290a381d1631.

PR #25, opened 2017-12-31
Strip annotations from initializer signatures

josiah-wolf-oberholtzer:
In `process_signature()`, the test `getattr(obj, '__annotations__', None)` fails for classes because it tests against the class object itself and not the initializer method. This results in type annotations not being stripped from the initializer signatures.

This PR just grabs the initializer method before continuing with the test for the presence of annotations.


josiah-wolf-oberholtzer:
Looks like Travis is failing due to the `backports.typing` installation, not because of the changes here.

agronholm:
Try rebasing your code on the latest master changeset.

agronholm:
You should also add a test that proves your patch actually fixes the problem.

agronholm:
Why did you close the PR?

josiah-wolf-oberholtzer:
That's odd. I didn't close it manually.

josiah-wolf-oberholtzer:
@agronholm I've implemented explicit testing of Sphinx's output (from the text builder, for simplicity) via the `sphinx-testing` package.

Changes include:
- Adding `sphinx-testing >= 0.7.2` to the test dependencies (https://github.com/sphinx-doc/sphinx-testing)
- Moving test collateral into a dedicated `test/` directory
- Adding stub `conf.py`, `index.rst` and `dummy_module.py` files to simulate a real Sphinx installation
- Adding a test of the actual Sphinx autodoc output for the objects defined in `dummy_module.py` (see `test_sphinx_output` in the test module)

The class defined in `dummy_module.py` has normal, class and static methods. You can see in the test output that the first argument to the class and static methods is being stripped, as reported in #23. If you're happy with the way I've reorganized the tests, I can provide a fix for those missing arguments.

josiah-wolf-oberholtzer:
Additionally, Travis is having site-wide build problems at the moment, hence the CI failures. Tests work fine locally.

josiah-wolf-oberholtzer:
I've just gone ahead and added a fix and tests for #23 via e40d296.

josiah-wolf-oberholtzer:
I've just found out about `sphinx.testing.fixtures`. It looks like those testing affordances supersede the older `sphinx-testing` package. I'll rework the test I added here to make use.

agronholm:
Looks pretty good overall. Please do rename the directory to `tests` though, as this is my standard naming. There are some code changes that I would like to make too but getting the test suite to pass is the first priority.

josiah-wolf-oberholtzer:
@agronholm Done.

agronholm:
Thanks, I'll take another look at this later today.

tests/test_sphinx_autodoc_typehints.py
```diff
@@ -96,3 +99,106 @@ def test_process_docstring_slot_wrapper():
...
+
+    app.build()
+
+    assert 'build succeeded' in status.getvalue()  # Build succeeded
+    assert not warning.getvalue().strip()  # No warnings
+
+    text_path = pathlib.Path(app.srcdir) / '_build' / 'text' / 'index.txt'
+    with text_path.open('r') as f:
+        text_contents = f.read().replace('–', '--')
+        text_contents == textwrap.dedent('''\
```
agronholm: Looks like this is missing an assert...?

tests/test_sphinx_autodoc_typehints.py
```diff
@@ -1,10 +1,13 @@
+import pathlib
 import pytest
 import sys
+import textwrap
 from typing import (
     Any, AnyStr, Callable, Dict, Generic, Mapping, Optional, Pattern, Tuple, TypeVar, Union)
 
 from sphinx_autodoc_typehints import format_annotation, process_docstring
 
+
```
agronholm: Just FYI, the try...except import block is the fourth group of imports – no extra empty lines!

agronholm:
It has some rough edges but it's good enough for merging and it solves a big practical problem.

agronholm:
Thanks!

josiah-wolf-oberholtzer:
Thanks for the merge. It's a start! This should fix #23 and #26.

agronholm:
I took a better look at the tests you provided. The output you test against does not appear to contain any of the alterations done by sphinx-autodoc-typehints. In fact, the plugin code does not appear to run at all. I can just raise an exception right at the top of `process_signature()` and the test still happily passes.

josiah-wolf-oberholtzer:
See #29.

The extension is running, but local changes are not visible after the initial test run. This is due to a bug in Sphinx's test fixtures.

PR #27, opened 2018-01-04
Avoid adding rtype directive if specified after return

auscompgeek:
The usual recommendation of ordering the return and rtype directives is
the return directive, then the rtype directive.

Previously, if users followed this recommendation, then an additional
rtype directive would be added before the return directive, resulting in
the return type showing twice in the resulting docs.


agronholm:
Thanks.

PR #29, opened 2018-01-16
Remove dummy project from temp on each test run.

josiah-wolf-oberholtzer:
This PR guarantees that the dummy project is removed from the temporary directory area at the start of each test tun.

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -1,5 +1,7 @@
 import os
 import pytest
+import pathlib
+import shutil
 from sphinx.testing.path import path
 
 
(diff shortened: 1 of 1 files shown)


agronholm:
One oddity I happened to observe: why do you add `yield` to fixtures that don't have any teardown code?

josiah-wolf-oberholtzer:
Ah, I simply forgot to strip that out when I adapted the fixture from https://github.com/pauleveritt/kaybee_prefactor/blob/5ee8adf8a5340e5c149b4a9acb3274a621cdc4bd/tests/integration/conftest.py#L21.

You can delete that statement.

PR #30, opened 2018-01-17
Fix for :rtype: being rendered in text

bmerry:
Adds a blank line before :rtype: when it is appended. This ensures that
Sphinx doesn't interpret it as part of normal text.

I've only reproduced the problem when using sphinx.ext.napoleon (just
having it loaded is enough; one doesn't need to use numpydoc
formatting), so it's added to the test suite.

### DIFF
--- a/sphinx_autodoc_typehints.py
+++ b/sphinx_autodoc_typehints.py
@@ -167,6 +167,12 @@ def process_docstring(app, what, name, obj, options, lines):
                         insert_index = i
 
                 if insert_index is not None:
+                    if insert_index == len(lines):
+                        # Ensure that :rtype: doesn't get joined with a paragraph of text, which
+                        # prevents it being interpreted.
+                        lines.append('')
+                        insert_index += 1
+
                     lines.insert(insert_index, ':rtype: {}'.format(formatted_annotation))
             else:
                 searchfor = ':param {}:'.format(argname)
--- a/tests/roots/test-dummy/conf.py
+++ b/tests/roots/test-dummy/conf.py
@@ -10,5 +10,6 @@ master_doc = 'index'
 
 extensions = [
     'sphinx.ext.autodoc',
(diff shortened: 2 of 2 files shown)


bmerry:
I imagine it's not great that the test is now run with sphinx.ext.napoleon always (i.e. no coverage of what happens without it loaded). I don't know what you want to do about it; possibly pytest or tox can parametrise around it, but I have very little experience with either.

sphinx_autodoc_typehints.py
```diff
@@ -167,6 +167,12 @@ def process_docstring(app, what, name, obj, options, lines):
                         insert_index = i
 
                 if insert_index is not None:
+                    if insert_index == len(lines):
+                        # Ensure that :rtype: doesn't get joined with a
+                        # paragraph of text, which prevents it being
+                        # interpreted.
+                        lines.append('')
+                        insert_index += 1
```
agronholm: A couple minor code style changes: add an empty line after the inner "if" block and reduce the number of comment lines (they can be longer). Otherwise this looks good.

bmerry: Sure, I can make those changes tomorrow. What's your preferred wrap column for comments?

agronholm: 100 (so wraps after 99), like the rest of the code.

bmerry: Done.

agronholm:
Thanks.

adamtheturtle:
I think that this closes https://github.com/agronholm/sphinx-autodoc-typehints/issues/24.

@agronholm - I would very much appreciate a release which includes this. Please let me know if I can help towards that in any way.

agronholm:
@adamtheturtle didn't 1.2.4 fix it for you?

adamtheturtle:
Thank you for doing a release @agronholm . This is a great project!

PR #35, opened 2018-02-13
Fix tilde sign appearance in HTML

adarshp:
This fixes an issue where TypeVar names had a tilde prefix appearing in the HTML version of the documentation.


agronholm:
If the prefix is left out, how then will users be able to tell if the typevar is invariant, covariant or contravariant?

adarshp:
My bad, I didn't realize the purpose of the tilde.

PR #36, opened 2018-03-15
fixed rendering of generic types outside of the typing module

tpoterba:
See the added test for clarity -- without this change, it renders as "B[]", instead of "B[int]".


agronholm:
I committed an alternate version on your behalf as 8fab2330f102bc209b12881008fc484e36fc4bbc. Thanks for the fix.

PR #43, opened 2018-06-01
Create MANIFEST.in

mariusvniekerk:
Ensure that we publish the license file with source dist packages on pypi


agronholm:
This project uses setuptools_scm which guarantees that all version controlled files are included in the sdist. The sdist available on PyPI contains the LICENSE file already, so what prompted you to create this PR?

PR #50, opened 2018-08-23
Fix exception when class has no documentation available

pleonex:
I am using a third-party ([ArtifactoryPath](https://github.com/devopshq/artifactory)) that doesn't provide documentation. this means that the Sphinx cannot add a link to the documentation but it should be able to print the name of the class.

After enabling this plugin, the following snippet of code was throwing the below exception because of the return type `ArtifactoryPath`. This PR fixes the issue by returning None if the class is not present in the dictionary.
```python
    def __get_artifactory_path(self, path: str) -> ArtifactoryPath:
        """Get a remote Artifactory path from a relative path.

        :param path: The relative URL to request.
        :returns: remote Artifactory path.
        """
        url = "{b}/{d}".format(b=ArtifactManager.BaseUrl, d=path)
```

Exception:
```
Exception occurred:
  File "/home/benito/Engineering/python-env/lib/python3.6/site-packages/sphinx_autodoc_typehints.py", line 150, in process_signature
    method_object = outer.__dict__[obj.__name__]
KeyError: '__get_artifactory_path'
```

### DIFF
--- a/sphinx_autodoc_typehints.py
+++ b/sphinx_autodoc_typehints.py
@@ -147,7 +147,7 @@ def process_signature(app, what: str, name: str, obj, options, signature, return
             outer = inspect.getmodule(obj)
             for clsname in obj.__qualname__.split('.')[:-1]:
                 outer = getattr(outer, clsname)
-            method_object = outer.__dict__[obj.__name__]
+            method_object = outer.__dict__.get(obj.__name__)
             if not isinstance(method_object, (classmethod, staticmethod)):
                 del argspec.args[0]
 


agronholm:
Would you mind adding a test to prevent any future regressions?

pleonex:
It turned out the problem isn't what I reported originally about external packages without documentation. After implementing the test I figure out the problem is with _dunders_ method because of the _mangling_. That is, it tries to search for `__my_method` when the name is `ClassName__my_method`.

I am working on the fix and tests.

agronholm:
I had a hunch this was the case :)

agronholm:
Those are not called "dunder" methods btw. Dunder methods are those with two underscores as both prefix and suffix (like `__repr__`). These are not mangled by Python.

pleonex:
Well, I think this is something the community never tends to agree. I learned the term from the book "Python Tricks" by Dan Bader, quoting:

```
Double underscores are often referred to as “dunders” in the Python community. The reason is that double underscores appear quite often in Python code, and to avoid fatiguing their jaw muscles, Pythonistas often shorten “double underscore” to “dunder.”

For example, you’d pronounce __baz as “dunder baz.” Likewise, __init__ would be pronounced as “dunder init,” even though one might think it should be “dunder init dunder.”
```

pleonex:
But that reminds me that there is a bug in my fix. I should check that the method doesn't end with "__" because in that case there isn't mangling. I'll fix that

PR #52, opened 2018-09-24
autodoc-process-signature chaining

jamesstidard:
I want to make some changes to the way my functions are rendered. Is it possible to register a call back like `autodoc-process-signature` to be called after `sphinx-autodoc-typehints` has finished its parsing?

Thanks.

### DIFF
--- a/sphinx_autodoc_typehints.py
+++ b/sphinx_autodoc_typehints.py
@@ -26,6 +26,10 @@ except ImportError:
         return func
 
 
+AUTODOC_TYPEHINTS_PROCCESS_SIGNATURE_EVENT = 'autodoc-typehints-process-signature'
+AUTODOC_TYPEHINTS_PROCCESS_DOCSTRING_EVENT = 'autodoc-typehints-process-docstring'
+
+
 def format_annotation(annotation):
     if inspect.isclass(annotation) and annotation.__module__ == 'builtins':
         if annotation.__qualname__ == 'NoneType':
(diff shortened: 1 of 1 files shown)


agronholm:
Hmm...okay, I see what you're getting at. Not sure that what you're suggesting is the right call. Isn't the normal way to do this to just add the extensions in the correct order in `conf.py`?

jamesstidard:
@agronholm yeah, I'm not familiar with the sphinx extension APIs or conventions - this was hacked together from reading yours and the sphinx source code.

Are you suggesting I make a custom extension and shim it in before sphine-autodoc-typehints?

My understanding, from reading the source code of autodoc, that it only calls `Sphinx.events.emit_firstresult` so then my extension would swallow the `autodoc-process-signature` event preventing `sphinx-autodoc-typehints` ever doing its thing.

I might be very muddled with my understanding though, maybe these events is not what I'm after. The actual think I'm trying to solve is I want to prevent some "private" arguments from being rendered in the documentation i.e:

```python
def my_func(ignore_me: str, doc_me: bool):
    ...
```

agronholm:
If you want to do postprocessing on the rendered docstring after sphinx-autodoc-typehints, then your sphinx extension should probably be registered after mine.

jamesstidard:
@agronholm , does that not leave me in the same position but the otherway round. If I register after yours then mine is never called as sphinx calls `Sphinx.events.emit_firstresult`. So only the first extension to register is allowed to do any post-processing.

agronholm:
This is starting to sound more and more like a Sphinx problem. Could you ask them for the preferred solution to this? Your original suggestion seems like a hack.

sphinx_autodoc_typehints.py
```diff
@@ -26,6 +26,10 @@ def _is_wrapper(f):
         return func
 
 
+AUTODOC_TYPEHINTS_PROCCESS_SIGNATURE_EVENT = 'autodoc-typehints-process-signature'
+AUTODOC_TYPEHINTS_PROCCESS_DOCSTRING_EVENT = 'autodoc-typehints-process-docstring'
```
eric-wieser: Typo: `PROCCESS` &rarr; PROCESS

sphinx_autodoc_typehints.py
```diff
@@ -160,7 +164,14 @@ def process_signature(app, what: str, name: str, obj, options, signature, return
             if not isinstance(method_object, (classmethod, staticmethod)):
                 del argspec.args[0]
 
-    return formatargspec(obj, *argspec[:-1]), None
+    signiture, return_annotation = formatargspec(obj, *argspec[:-1]), None
```
eric-wieser: typo: `signiture` &rarr; `signature`

Issue #51, opened 2018-08-28
formatargspec is deprecated


frankier:
It's called here https://github.com/agronholm/sphinx-autodoc-typehints/blob/38ff0554bcf2d30b1886002be24deff59d665072/sphinx_autodoc_typehints.py#L163

The warning happens in Sphinx 2.x. It looks like this:

    .../lib/python3.6/site-packages/sphinx/ext/autodoc/inspector.py:117: RemovedInSphinx20Warning: formatargspec() is now deprecated.  Please use sphinx.util.inspect.Signature instead.
      RemovedInSphinx20Warning)

agronholm:
Closed via 410958966cba3976444b0c1662409e5af000e21f.

PR #55, opened 2018-10-21
Fix Sphinx deprecation warning: use "Signature" instead of "formatargspec"

yoeo:
Fixes issue  #51 "formatargspec is deprecated"
All tests pass.


agronholm:
Thanks!

TiemenSch:
Would you consider increasing the version number for this?

agronholm:
One I make a new release, it will get a +0.1 increase in the version number.

PR #57, opened 2018-11-06
adding support for trailing underscores

### LINKED ISSUE #56: types not associated with parameters which have trailing underscores
parameters that have a trailing underscore need to be escaped in sphinx (see this issue: https://github.com/sphinx-doc/sphinx/issues/519)

so the documentation needs to work as follows:

```python
def factory(type_: type):
    """
    :param type\\_: a type
    :type type\\_: type
    """
```

when using this extension the aforementioned bug would appear to change this example:

```python
def factory(type_: type):
    """
    :param type\\_: a type
    """
```

into something equivalent to the following:

```python
def factory(type_: type):
    """
    :param type\\_: a type
    :type type_: type
    """
```

and as a result the type is not associated with the parameter and rendered in the output.

danielknell:
resolves #56

### DIFF
--- a/sphinx_autodoc_typehints.py
+++ b/sphinx_autodoc_typehints.py
@@ -183,6 +183,9 @@ def process_docstring(app, what, name, obj, options, lines):
             return
 
         for argname, annotation in type_hints.items():
+            if argname.endswith('_'):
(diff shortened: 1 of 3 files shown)


agronholm:
Thanks!

PR #61, opened 2018-11-19
using default annotation if get_type_hints fails

fariza:
Sometimes get_type_hints fails because it can not resolve forward references raising a NameError, it is better just to use the standard annotation that will be resolved as string in most cases.

This might happen for different reasons, among other things because the module was not imported.

### DIFF
--- a/sphinx_autodoc_typehints.py
+++ b/sphinx_autodoc_typehints.py
@@ -182,6 +182,8 @@ def process_docstring(app, what, name, obj, options, lines):
         except (AttributeError, TypeError):
             # Introspecting a slot wrapper will raise TypeError
             return
+        except NameError:
+            type_hints = obj.__annotations__
 
         for argname, annotation in type_hints.items():
             if argname.endswith('_'):
--- a/sphinx_autodoc_typehints.py
+++ b/sphinx_autodoc_typehints.py
@@ -3,6 +3,9 @@ import typing
 from typing import get_type_hints, TypeVar, Any, AnyStr, Generic, Union
 
 from sphinx.util.inspect import Signature
+from sphinx.util import logging
+
+logger = logging.getLogger(__name__)
 
 try:
     from inspect import unwrap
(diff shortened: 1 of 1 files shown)


fariza:
I am trying to use it in an already documented code, that runs fine with sphinx, only the annotations are ugly as they are displayed in the signature (as you know).

So when I add this extension, it fails with NameError without even mentionning the line where the error ocurred. 
It is basically an exception that you didn't catch in the try/except block

In this particular case, that exception comes from the module not being imported.
For example
`/repo/package1/module1/file1.py`
`/repo/package1/module2/file2.py`

Inside `file1` there is `Class1` 
Inside `file2` there is `Class2` with a signature `somemethod(self, var: 'module1.file1.Class1')` 

If in `file2` I don't import module1, it will raise the NameError

The point of not importing the module is because it was already like that and should work, even if the output is not perfect (it will be strings), still the output will be there

agronholm:
Generally the rule in type annotations is that they should be valid. Thus you will want to import them, at least conditionally (`if typing.TYPE_CHECKING: import module1.file1`).

fariza:
Yes I agree.
1) those are valid forward references, even if the version of typing doesn't fully support them (python < 3.7)
2) It was working before, even if the output wasn't great, it did generate the documentation, now it doesn't run.
3) If for some reason it can not get the proper annotations (bad annotations), it should at least use the string as fallback. Not fail

agronholm:
A forward reference is an arbitrary expression that can be evaluated successfully when the module has been fully imported. In this case they cannot be since you have not imported the modules they refer to. What version of sphinx-autodoc-typehints did this work with before?

fariza:
I mean before using the extension.

fariza:
Maybe emit warnings but not fail

agronholm:
Ok, well can you fix the forward references by doing the conditional imports? AFAIK mypy checks would also fail here for the same reason.

agronholm:
I will likely do as you suggested and make it emit some warning.

fariza:
I'll update the PR with a warning and see

fariza:
Before it just stopped:
```
....
Exception occurred:
  File "<string>", line 1, in <module>
NameError: name 'feature' is not defined
The full traceback has been saved in /tmp/sphinx-err-1ureskfv.log, if you want to report the issue to the developers.
```

Now it raises a warning:
```
....
WARNING: Impossible to ge_type_hints for axviewer.full_reference_wrapper.FSIM.__init__ (name 'feature' is not defined)
....
```

fariza:
@agronholm is there any case where the `AttributeError` or `TypeError` are rised but the `__annotations__` attribute still there?
I was just wondering if it will be more general grouping the three errors together and trying to get the `__annotations__` string

PR #62, opened 2018-11-20
Bump up minimum Sphinx version (fixes #59)

### LINKED ISSUE #59: Version 1.5.0 breaks with Sphinx<1.8
The event `config-inited` was added on Sphinx 1.8 (https://www.sphinx-doc.org/en/master/extdev/appapi.html#event-config-inited) and it's used by [sphinx_autodoc_typehints.py](https://github.com/agronholm/sphinx-autodoc-typehints/blob/master/sphinx_autodoc_typehints.py#L228).

However, sphinx_autodoc_typehints requires only [Sphinx>=1.7](https://github.com/agronholm/sphinx-autodoc-typehints/blob/master/setup.cfg#L29).

Therefore, sphinx-autodoc-typehints 1.5.0 breaks with Sphinx<1.8, which is forced by e.g. `readthedocs.org`.

Should we bump the required Sphinx version or avoid the `config-inited` event?

See also https://github.com/agronholm/sphinx-autodoc-typehints/commit/2afccd39f9e559f4cbd593461080cdfede7c2d50#r31313681 for my inline comment.

michelole:
As `sphinx_autodoc_typehints` now uses `config-inited` events introduced in Sphinx 1.8 (https://www.sphinx-doc.org/en/master/extdev/appapi.html#event-config-inited), bump up the minimum required Sphinx version to 1.8.

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -26,7 +26,7 @@ classifiers =
 py_modules = sphinx_autodoc_typehints
 python_requires = !=3.5.0, !=3.5.1
 install_requires =
-    Sphinx >= 1.7
+    Sphinx >= 1.8
     typing >= 3.5; python_version == "3.4"
 
 [options.extras_require]


agronholm:
I used a different event instead so this will not be necessary.

PR #64, opened 2018-11-21
Identify typing.NewType wrappers and construct annotation references

### LINKED ISSUE #41: Types created with NewType are rendered poorly.
If I have code like:
```
    Foo = typing.NewType('Foo', str)
    def bar(x: Foo):
        pass
```
then its parameter gets documented as:
```
    x (<function NewType.<locals>.new_type at 0x7f644b4c6488>)
```
I was expecting something like one of the following:
```
    x (Foo)
    x (Foo alias for str)
    x (str)
```

gwax:
Resolves #41 producing output of the form `NewType(NewName, OldType)` with appropriate references.

### DIFF
--- a/sphinx_autodoc_typehints.py
+++ b/sphinx_autodoc_typehints.py
@@ -119,8 +119,8 @@ def format_annotation(annotation):
         return '...'
     elif (inspect.isfunction(annotation) and annotation.__module__ == 'typing' and
           hasattr(annotation, '__name__') and hasattr(annotation, '__supertype__')):
-        # NewType
-        return ':py:func:`~typing.NewType`\\(:py:data:`~{}`, {})'.format(annotation.__name__, format_annotation(annotation.__supertype__))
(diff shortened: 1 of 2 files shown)


rmorshea:
@agronholm thoughts on this?

agronholm:
I'll try to take a look at this this weekend.

agronholm:
Thanks!

agronholm:
@gwax I've been rechecking the code and I cannot figure out why you chose to format the name of the NewType as `:py:data:~TypeName`. Shouldn't it contain the module name as well, like `:py:data:~module.name.TypeName`? I'm also wondering if the formatting scheme is even in the right direction. Maybe just `:py:data:~module.name.TypeName` should be used, without involving `typing.NewType`?

agronholm:
Ok, the module name containing the new type is not recorded anywhere so this is the best we can do. Sorry for the noise :)

PR #71, opened 2019-01-25
Allow fully qualified class names

Hwesta:
When rendering the type hints, add a flag to allow them to always be displayed fully qualified (module + class name).

I have a lot of type hints that belong to external modules, like `aiohttp.Request` or `databaseprovder.Session` and it's much easier to tell what's going on when they render as `aiohttp.Request` instead of `Request`.

Since not everyone may want this, I added a new flag `typehints_fully_qualified` to enable this behaviour.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -57,7 +57,9 @@ The following configuration options are accepted:
 
 * ``set_type_checking_flag`` (default: ``True``): if ``True``, set ``typing.TYPE_CHECKING`` to
   ``True`` to enable "expensive" typing imports
-* ``typehints_fully_qualified`` (default: ``False``): if ``True``, class names are always fully qualified (e.g. ``module.for.Class``). If ``False``, just the class name displays (e.g. ``Class``)
+* ``typehints_fully_qualified`` (default: ``False``): if ``True``, class names are always fully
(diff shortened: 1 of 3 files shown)


agronholm:
Sorry for not getting to this earlier. I have one question: why is the option's `rebuild` parameter set as `env`? This should only affect the rendering of the document and not require a full environment rebuild.

Hwesta:
With the rebuild parameter as `html` when I changed the value and rebuilt the docs with `make html` it didn't notice the updated parameter and didn't rebuild the docs.  By setting it to `env` when I changed it and tried to rebuild the docs it saw the change and regenerated the HTML like I expected.  It does seem like more of a `html`-level config, but I was frustrated that changing it didn't rebuild the docs. :woman_shrugging: 

I noticed the README for the `set_type_checking_flag` default is wrong, so I added a fix to this branch as well.

agronholm:
Sorry again for taking so long with this, but would you mind adding at least one test to the test suite for this feature? I'll merge this when you do that.

agronholm:
Thank you, and sorry about the long wait!

Hwesta:
Thanks for maintaining this project and working with me on this PR! Maintainership is often a thankless job, and I appreciate this package and the work you put into keeping it up.

Issue #31, opened 2018-01-17
v1.2.4 causing AttributeErrors


goodspark:
```
AttributeError: module 'my.module' has no attribute 'accept.<locals>'
```

Does not occur in v1.2.3

It appears to happen whenever I use [hypothesis] in a test class.

```python
from hypothesis import given, integers


class TestSomething:
  @given(seconds=integers(min_value=1, max_value=100))
  def test_something(self, seconds: int):
    pass


# does not repro if no test class is used
@given(seconds=integers(min_value=1, max_value=100))
def test_something_else(seconds: int):
  pass
```

[hypothesis]: http://hypothesis.readthedocs.io/

<details>
<summary>Sphinx log output</summary>

```
# Sphinx version: 1.6.5
# Python version: 3.6.3 (CPython)
# Docutils version: 0.14 
# Jinja2 version: 2.10
# Last messages:
#   reading sources... [ 19%] ref/tests.plugin.pyramid
#   
#   reading sources... [ 20%] ref/tests.plugin.settings_fixtures
#   
#   reading sources... [ 21%] ref/tests.project
#   
#   reading sources... [ 22%] ref/tests.project.models
#   
#   reading sources... [ 24%] ref/tests.project.models.test_project
#   
# Loaded extensions:
#   alabaster (0.7.10) from /home/my/code/project/.tox/docs/lib/python3.6/site-packages/alabaster/__init__.py
#   sphinx.ext.autodoc (1.6.5) from /home/my/code/project/.tox/docs/lib/python3.6/site-packages/sphinx/ext/autodoc.py
#   sphinx.ext.intersphinx (1.6.5) from /home/my/code/project/.tox/docs/lib/python3.6/site-packages/sphinx/ext/intersphinx.py
#   sphinx.ext.viewcode (1.6.5) from /home/my/code/project/.tox/docs/lib/python3.6/site-packages/sphinx/ext/viewcode.py
#   sphinx.ext.extlinks (1.6.5) from /home/my/code/project/.tox/docs/lib/python3.6/site-packages/sphinx/ext/extlinks.py
#   sphinx_autodoc_typehints (unknown version) from /home/my/code/project/.tox/docs/lib/python3.6/site-packages/sphinx_autodoc_typehints.py
#   ghlink (unknown version) from /home/my/code/project/docs/_extensions/ghlink.py
[...]

PR #74, opened 2019-02-25
Ignore callables defined as local functions

kmyk:
a pull request for #31

When such wrong callables found, this ignore them and prints a warning which recommends `functools.wraps`.


agronholm:
Thanks. Sorry this took a while!

PR #78, opened 2019-04-26
Allow users to strip annotations from signatures without modifying docstrings

goodmami:
Some users may wish to use type hints in their code for purposes other than Sphinx documentation, and sphinx-autodoc-typehints is useful for stripping the annotations out of the signatures when autodoc is used, but there was no way to prevent sphinx-autodoc-typehints from also modifying the docstrings. This PR adds a new option `add_type_hints_to_docstring` which defaults to `True` (thus not changing the current behavior) but can be set to `False` to prevent any modification to the docstrings.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -58,6 +58,9 @@ The following configuration options are accepted:
 * ``set_type_checking_flag`` (default: ``False``): if ``True``, set ``typing.TYPE_CHECKING`` to
   ``True`` to enable "expensive" typing imports
 
+* ``add_type_hints_to_docstring`` (default: ``True``): if ``True``, insert parameter types from the
+  signatures of callable objects into their docstrings
+
 
 How it works
 ------------
--- a/README.rst
+++ b/README.rst
@@ -55,7 +55,7 @@ Options
 
 The following configuration options are accepted:
 
-* ``set_type_checking_flag`` (default: ``True``): if ``True``, set ``typing.TYPE_CHECKING`` to
+* ``set_type_checking_flag`` (default: ``False``): if ``True``, set ``typing.TYPE_CHECKING`` to
   ``True`` to enable "expensive" typing imports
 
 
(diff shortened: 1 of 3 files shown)


goodmami:
In the meantime (or if this PR gets rejected), adding the following hack to Sphinx's `conf.py` will accomplish the same thing as setting `add_type_hints_to_docstring = False`:

```python
# Use sphinx-autodoc-typehints to strip annotations from signatures
# but don't insert types into docstrings
import sphinx_autodoc_typehints

def noop_process_docstring(app, what, name, obj, options, lines):
    pass

sphinx_autodoc_typehints.process_docstring = noop_process_docstring
```

untitaker:
I am confused about this PR. To me the entire purpose of this plugin is to insert type hints into docstrings. If I set this flag to non-default, what functionality is left?

goodmami:
> If I set this flag to non-default, what functionality is left?

Please see the other half of this PR's title: "Allow users to strip annotations from signatures"

eric-wieser:
Sphinx has a builtin [`autodoc_typehints`](http://www.sphinx-doc.org/en/master/usage/extensions/autodoc.html#confval-autodoc_typehints) configuration setting - can this be piggy-backed on?

goodmami:
@eric-wieser thanks, I hadn't noticed that option. In principle I could use that for my use-case, however when I tried it I found the results are not as good as sphinx-autodoc-typehints:

* it does not remove return-type annotations
* it inserts whitespace when an annotated parameter has a default (`param: bool = False` becomes *param = False* whereas `param=False` becomes *param=False*)

These issues should be fixable, though.

goodmami:
The Sphinx-doc issue I created (sphinx-doc/sphinx#6589) to improve the formatting of stripped annotations has been resolved and the fix will be available in v2.2.0. I think this PR can now be closed (not merged). I'll do that now.

Thanks @eric-wieser for pointing out the `autodoc_typehints` built-in option.

A couple notes:

* Users may arrive here (as I did) without first discovering the built-in option; it may be useful for the README to point to the documentation for that.
* This PR also fixed the default value of `set_type_checking_flag` in the README, but this has now been resolved by a separate commit, so there's no risk of it being lost.

PR #87, opened 2019-07-03
feat: Add type hint comment support

### LINKED ISSUE #13: type hint comment support
this would allow using it under Python 2.7 :+1:

untitaker:
Fix #13

I've made it so that there's as little risk of regression as possible for existing users by shortcutting to `get_type_hints` or `__annotations__` if that approach works. If those return empty mappings, use `typed_ast` with `inspect.getsource` to get type hint comments. An additional advantage is that `typed_ast` is only an optional dependency just for type hint comments.

### DIFF
--- a/sphinx_autodoc_typehints.py
+++ b/sphinx_autodoc_typehints.py
@@ -272,7 +272,7 @@ def backfill_type_hints(obj, name):
     if comment_returns:
         rv['return'] = comment_returns
 
-    if comment_args_str != '(...)':
+    if comment_args_str not in ('()', '(...)'):
         logger.warning(
             'Only supporting `type: (...) -> rv`-style type hint comments, '
             'skipping types for "%s"',
--- a/sphinx_autodoc_typehints.py
+++ b/sphinx_autodoc_typehints.py
@@ -250,9 +250,8 @@ def get_all_type_hints(obj, name):
 
     if comment_args == ['...']:
         comment_args = [None] * len(args)
-    elif len(comment_args) != len(args):
-        logger.warning('Wrong argument count in type hint comment for "%s"', name)
-        return rv
+    elif len(args) > len(comment_args):
+        comment_args = [None] * (len(args) - len(comment_args)) + comment_args
 
     rv['return'] = comment_returns
 
--- a/sphinx_autodoc_typehints.py
+++ b/sphinx_autodoc_typehints.py
@@ -191,6 +191,8 @@ def process_signature(app, what: str, name: str, obj, options, signature, return
 
 
 def get_all_type_hints(obj, name):
+    rv = {}
+
     try:
         rv = get_type_hints(obj)
     except (AttributeError, TypeError):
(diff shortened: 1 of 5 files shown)


tests/test_sphinx_autodoc_typehints.py
```diff
@@ -296,4 +296,28 @@ class InnerClass
...
+
+           Return type:
+              None
+
+        class dummy_module.ClassWithTypehints(x)
+
+           Class docstring.
+
+           Parameters:
+              **x** -- foo
```
untitaker: I know this looks like any other unresolvable type. I think we could fix that in the least lines of code by *setting* `__annotations__` on the obj, and calling `get_type_hints` again to resolve the names we extracted from the type hint comments... lmk what you think of this approach

agronholm:
Aside from some minor code formatting nits, this looks good to me. Thanks!

untitaker:
@agronholm feel free to point those nits out

agronholm:
Parameters put on too many lines, and backticks used where a double quote would have been preferred by me. Nothing worth fussing about, really.

eric-wieser:
It would be fantastic if this could be rolled into a PyPI release

eric-wieser:
xref sphinx-doc/sphinx#2755

agronholm:
I will do that soon once I'm done with the outstanding patches lined up for that release.

agronholm:
This is now in v1.7.0.

sphinx_autodoc_typehints.py
```diff
@@ -189,6 +190,115 @@ def process_signature(app, what: str, name: str, obj, options, signature, return
...
+    try:
+        import typed_ast.ast3 as ast
+    except ImportError:
+        return rv
+
+    def _one_child(module):
+        children = list(ast.iter_child_nodes(module))
+
+        if len(children) != 1:
+            logger.warning(
```
gaborbernat: Why this check? This basically disallows doing:
```
def magic(router, **options):
# type: (str, Any) -> Callable[[Any], Any]
```

untitaker: Is this indentation intended? I ran into this warning a bunch of times but didn't investigate which items produce two ast nodes. I am also not sure what to do with two ast nodes

gaborbernat: just select the first function definition? (type ignore elements should be definitely filtered out) - with the current implementation if the body has type ignores in it this will fail it

agronholm: And what about `@override`?

gaborbernat: can probably come up with something sane 🤔

untitaker: As said, this doesn't handle all cases, but bailing out seemed safer than selecting a random node without knowledge of why it yielded multiple.

decorators seemed to work for me, but it would make sense to get multiple nodes for that.

gaborbernat: created https://github.com/agronholm/sphinx-autodoc-typehints/pull/100/files that adds non inlined type comment support 😃

PR #92, opened 2019-08-06
Catch recursion errors when getting type hints

### LINKED ISSUE #91: Recursion errors with some recursive types
This type definition

```python
TreeType = typing.Dict[str, typing.Union['TreeType', typing.List['TreeType']]]
```

causes documentation generation to fail with the following recursion error.

```
Recursion error:
maximum recursion depth exceeded while calling a Python object
```

It appears that this only happens when the recursive type is referenced twice (*e.g* `TreeType = typing.Dict[str, 'TreeType']` works fine).

sangaline:
There's an [upstream issue](https://github.com/python/typing/issues/574) which is causing documentation building to fail when recursive type definitions are used. This catches `RecursionError` exceptions when calling `typing.get_type_hints()` so that the build can proceed.

Closes #91

### DIFF
--- a/sphinx_autodoc_typehints.py
+++ b/sphinx_autodoc_typehints.py
@@ -191,8 +191,9 @@ def get_all_type_hints(obj, name):
 
     try:
         rv = get_type_hints(obj)
-    except (AttributeError, TypeError):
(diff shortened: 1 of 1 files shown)


agronholm:
Thanks.

PR #95, opened 2019-08-21
Create correct links for more (all?) typing.* types.

### LINKED ISSUE #94: Support all `typing.*` types listed as :py:data:
As you know, some types aren’t `py:class`es but `py:data`. Currently handled are `Any`, `AnyStr`, `Callable`, `Optional` and `Union`, but there are more, most importantly `Tuple`:

```console
$ python -m sphinx.ext.intersphinx https://docs.python.org/3/objects.inv | grep -P '^([^\s]|\s+typing\.).*'
...
py:data
#       typing.Any                               library/typing.html#typing.Any
#       typing.AnyStr                            library/typing.html#typing.AnyStr
#       typing.Callable                          library/typing.html#typing.Callable
        typing.ClassVar                          library/typing.html#typing.ClassVar
        typing.NoReturn                          library/typing.html#typing.NoReturn
#       typing.Optional                          library/typing.html#typing.Optional
        typing.Tuple                             library/typing.html#typing.Tuple
#       typing.Union                             library/typing.html#typing.Union
...
```

### LINKED ISSUE #96: typing.TextIO gets mangled
We call `annotation_cls.__qualname__title()` to make e.g. `'tuple'` into `'Tuple'`, but this breaks `TextIO`, which becomes mangled into `'Textio'`.

flying-sheep:
faa272c fixes #94, types that should be linked like ``:py:data:`typing.Tuple` `` instead of  `py:class`.

1f5e310 fixes #96, only `.title()`casing types that are builtins, such as `set`→`Set` and `dict`→`Dict`, but not `TextIO`→`Textio`.

### DIFF
--- a/sphinx_autodoc_typehints.py
+++ b/sphinx_autodoc_typehints.py
@@ -1,3 +1,4 @@
+import builtins
 import inspect
 import textwrap
 import typing
@@ -76,14 +77,17 @@ def format_annotation(annotation, fully_qualified=False):
             class_name = 'ClassVar'
             if hasattr(annotation, '__args__'):
                 params = annotation.__args__
-            elif getattr(annotation, '__type__', None):
+            elif getattr(annotation, '__type__', None) is not None:
                 params = [annotation.__type__]
-        # Python 3.5’s typing.Tuple. 3.6’s has .__origin__=tuple and .__args__=[...]
-        elif annotation_cls.__qualname__ == 'Tuple' and hasattr(annotation, '__tuple_params__'):
+        elif annotation_cls.__qualname__.lower() == 'tuple':
             prefix = ':py:data:'
-            params = annotation.__tuple_params__
-            if annotation.__tuple_use_ellipsis__:
-                params += (Ellipsis,)
+            class_name = 'Tuple'
+            if hasattr(annotation, '__args__'):
+                params = annotation.__args__
+            elif hasattr(annotation, '__tuple_params__'):
+                params = annotation.__tuple_params__
+                if annotation.__tuple_use_ellipsis__:
+                    params += (Ellipsis,)
         elif annotation_cls.__qualname__ == 'Callable':
             prefix = ':py:data:'
             arg_annotations = result_annotation = None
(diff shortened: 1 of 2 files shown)


flying-sheep:
I can’t say I understand the test failure. For some reason the `Parameters` and `Result` section are sometimes switched. I had that locally before my last commit, but not anymore, here it’s the other way around. I’m pretty sure it’s unrelated to my changes.

agronholm:
If you run the test suite for py35 using `tox` multiple times, do you get consistent results?

agronholm:
Also, `typing.NoReturn` was only introduced in Python 3.5.4 while this library supports 3.5.2+ (due to the fact that 3.5.2 is still the latest version of Python 3.5 on Ubuntu 16.04 LTS). Therefore it needs to be imported conditionally. The same applies to `typing.ClassVar` which was introduced in 3.5.3.

flying-sheep:
> If you run the test suite for py35 using `tox` multiple times, do you get consistent results?

Yes, but the inverse results from travis: For me it failed consistently after the first commit and succeeded consistently after the second one. (with the same error as travis)

agronholm:
I'll try to review this before the end of the week. I've been busy with many projects lately.

flying-sheep:
Hi! What’s up?

agronholm:
Been busy getting a new release of AnyIO out the door. I will get back to this project after that.

agronholm:
Okay, with that now out of the way I can focus on this. I'm currently looking for a better option for getting proper names out of the annotations rather than using either special casing or `.title()`.

flying-sheep:
Great, but this is still an improvement, and already gets some tests in place, so why not merge this first?

agronholm:
I've committed my own version of the title case/data fix to master. Is it enough or did I miss something? If I didn't, then I will close this PR. Thank you for bringing this up either way.

agronholm:
Seems I missed `Tuple`. It's really frustrating to not be able to tell when something is a class and when it's data. Seems wildly inconsistent.

agronholm:
Also, I realize you made this PR on my request, but I decided that I wanted to simplify the code instead. Apologies for that :(

flying-sheep:
It’s OK!

PR #98, opened 2019-09-09
Change detection if a typing.* item is py:data

### LINKED ISSUE #97: Test correct classification of typing exports
You fixed `py:data` links to typing in 3c9f4f38fce0003a102e0436b6b34f85829dab48, but expressed frustration that there isn’t an easy way to get which things are `py:data` and which `py:class`. We could try to devise a method, and test it by loading Python’s objects.inv via `sphobjinv` and comparing our results vs the ones in a given Python version.

flying-sheep:
I only found an elegant way for plain types, i.e. ` AnyStr` (a `TypeVar`), `Any`/`NoReturn` (`_SpecialForm`s), and the non-subset versions of `Callable`/`Tuple` (`_VariadicGenericAlias`es) and `ClassVar`/`Optional`/`Union` (`_SpecialForm`s). As soon as you subset them, they all become `_GenericAlias`es:

```py
>>> {a: type(a) for a in [Callable[[], None], Tuple[str, int], ClassVar[str], Union[str, int]]}
{typing.Callable[[], NoneType]: typing._GenericAlias,
 typing.Tuple[str, int]: typing._GenericAlias,
 typing.ClassVar[str]: typing._GenericAlias,
 typing.Union[str, int]: typing._GenericAlias}
```

Sadly, other things become `_GenericAlias`es too, e.g. `type(typing.Generator[None, None, None])` → `typing._GenericAlias`, so we need to go over the information stored in the `_GenericAlias`: `_inst`, `_special`, `_name`, `__origin__`, and `__args__`. Of those, only `__origin__` and `_name` are of interest. `_name` is derived from `__origin__` using `typing._normalize_alias`, which maps `tuple`→`Tuple`, `defaultdict`→`DefaultDict` for display purposes.

Fixes #97

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -33,6 +33,7 @@ test =
     typing_extensions >= 3.5
     typed_ast >= 1.4.0
     dataclasses; python_version == "3.6"
+    sphobjinv >= 2.0
 
 [flake8]
 max-line-length = 99
--- a/sphinx_autodoc_typehints.py
+++ b/sphinx_autodoc_typehints.py
@@ -1,5 +1,6 @@
 import inspect
 import textwrap
+import collections.abc
 import typing
 from typing import get_type_hints, TypeVar, Any, AnyStr, Generic, Union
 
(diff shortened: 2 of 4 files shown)


flying-sheep:
Hmm, on the other hand, this obviously depends on Python 3.7+ only private classes so probably not ideal -.-

agronholm:
Yeah, sorry but I don't see this as an improvement. Thanks for trying though.

flying-sheep:
Actually the testing part is. Could you reopen? I’ll get rid of the detection changes and we can merge just additional tests.

flying-sheep:
OK, this should do it!

agronholm:
I'm not ready to merge this yet. Something needs to be done to prevent it from downloading the inventory on every tox run on the same machine. I'd also prefer it to only fetch the inventory of the Python version it's running on.

flying-sheep:
OK, that’s actually nice and easy!

flying-sheep:
Hmm, 306b856a856cd71baf9be6271dd53ab2ce653386 didn’t fix the slow runtime of the test, why?

agronholm:
The fixture is still function scoped and not session scoped, could that be the problem?

PR #100, opened 2019-09-11
Support type comments that are not inlined

gaborbernat:


agronholm:
What is the benefit of this?

agronholm:
Clarification: that comment was directed at the changes made to `setup.cfg` and `tox.ini`.

gaborbernat:
@agronholm the benefit is that build backends no longer need to communicate via ``setup.py`` invocations for packaging but have a well defined programming API via https://www.python.org/dev/peps/pep-0517/#build-wheel. It's the new best practice in the world of Python. More in-detail at https://www.bernat.tech/pep-517-518/

setup.cfg
```diff
@@ -31,8 +31,9 @@ install_requires = Sphinx >= 2.1
 test =
     pytest >= 3.1.0
     typing_extensions >= 3.5
-    typed_ast >= 1.4.0; python_version < "3.8"
     dataclasses; python_version == "3.6"
+comment =
```
gaborbernat: this allows users to enable the comment support via ``sphinx-autodoc-typehints[comments]`` rather than have to know the typed ast library by name

gaborbernat:
@agronholm would be **much-much** appreciated a release of this today if it looks good otherwise 👍

agronholm:
Those changes have nothing whatsoever to do with comment inlining, so if you're going to push for that I'd appreciate if you did so in a separate PR.

gaborbernat:
is under https://github.com/agronholm/sphinx-autodoc-typehints/pull/101 👍

PR #101, opened 2019-09-11
add PEP-517 support

gaborbernat:


pyproject.toml
```diff
@@ -1,2 +1,7 @@
 [build-system]
-requires = ["setuptools >= 36.2.7", "wheel", "setuptools_scm >= 1.7.0"]
+requires = [
+    "setuptools >= 40.0.4",
+    "setuptools_scm >= 2.0.0, <4",
```
agronholm: Why the added constraint here? Is there something happening with setuptools_scm at version 4?
You also upgraded the minimum version to 2.0.0, any reason? Not that I'm against it.

tox.ini
```diff
@@ -2,9 +2,10 @@
 minversion = 2.5.0
```
agronholm: This needs to be upgraded to 3.3.0, yes?

setup.cfg
```diff
@@ -31,8 +31,9 @@ install_requires = Sphinx >= 2.1
 test =
     pytest >= 3.1.0
     typing_extensions >= 3.5
-    typed_ast >= 1.4.0; python_version < "3.8"
     dataclasses; python_version == "3.6"
+comment =
```
agronholm: I would prefer something more descriptive here, say `type_comments`.
Also, please add a mention of this in the README.

pyproject.toml
gaborbernat: Less than four so the packaging does not fail with release with 4.0. Once it comes out and we validate it will not we can bump. Had some issues in the past with version 1.0 so bumped to 2.0. Don't remember exactly.

tox.ini
gaborbernat: Indeed.

setup.cfg
gaborbernat: Fixed.

pyproject.toml
agronholm: I'm okay with bumping the requirement to 2.0 but I prefer being optimistic so please remove the `<4` constraint. It's not a huge deal either way as the wheels will still remain installable. If something breaks with v4 then we will deal with that then.

gaborbernat: But that would render all packages released now unusable. Semantic versioning is not designed for optimistic view of major version breakage...

agronholm: Huh? How so? `pyproject.toml` is not touched when the wheel is installed, which is how virtually everybody will be installing this library. So even if setuptools_scm did break packaging for this library, it would not affect most people immediately.

gaborbernat: It would effect anyone using the source distribution directly, which is the case with a lot of companies. A lot of enterprise environments prefer to build everything from source for transparency 🤔

gaborbernat: there's almost no benefit to not pinning 🤔

agronholm: Then why are you singling out setuptools_scm? Why not pin setuptools and wheel as well?

gaborbernat: wheel should be; setuptools is highly bound at this point, and they use semantic versioning only in name 👍

agronholm: From what I've seen, Jason has been trying to be really strict about semver compliance with setuptools, which is why it has so ridiculously many major version bumps in its history. As for wheel, I have not even started the 1.0 series so it actually would not do any good to bind it at this point.

gaborbernat: Well once they do an breaks protocol of 0.x it would be useful. As for setuptools I suppose it's all about what parts of it one uses. There's no guarantee next major will not break exactly the part we're using so could be pinned.

agronholm: Besides, enterprise customers tend to pin their dependencies to exact versions, so setuptools_scm breaking the install will only be a problem when they try to upgrade. Plus, as a workaround they can always allow the wheel as an exception, given that it's pure Python anyway.

agronholm: Please remove the pins. I will merge this when you have done that.

agronholm:
Thanks.

PR #102, opened 2019-09-11
Default values mess up type comment argument detection

gaborbernat:
This fixes it by collecting them in the correct order. Validated with my project of 250+ type annotated functions.

### DIFF
--- a/sphinx_autodoc_typehints.py
+++ b/sphinx_autodoc_typehints.py
@@ -223,14 +223,12 @@ def get_all_type_hints(obj, name):
 
 
 def backfill_type_hints(obj, name):
-    rv = {}
-
     parse_kwargs = {}
     if sys.version_info < (3, 8):
         try:
             import typed_ast.ast3 as ast
         except ImportError:
-            return rv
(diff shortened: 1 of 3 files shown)


agronholm:
Would it be possible to add a test for this?

gaborbernat:
Yes 👍 though the tests in here already cover it, by I assume you meant an example of where the current implementation fails 🤔

gaborbernat:
@agronholm can you review/merge/release it? thanks!

sphinx_autodoc_typehints.py
```diff
@@ -248,35 +246,31 @@ def _one_child(module):
     try:
         obj_ast = ast.parse(textwrap.dedent(inspect.getsource(obj)), **parse_kwargs)
     except TypeError:
-        return rv
+        return {}
 
     obj_ast = _one_child(obj_ast)
     if obj_ast is None:
-        return rv
+        return {}
```
gaborbernat: made these to make the code more readable, this way it's more explicit that we bail in these cases

PR #103, opened 2019-09-12
Add default to formatted annotation

### LINKED ISSUE #48: Also annotate parameter defaults.
In numpy docstrings I often see

```rst
Parameters
----------
myparam : `Type`, optional (default: `None`)
    My param does things.
```

I prefer styling the parameter docs as definition list like numpydoc, where there is enough space for type and defaults:

<blockquote>
<dl>
<dt><strong>myparam</strong>: <code>Type</code>, optional (default: <code>None</code>)</dt>
<dd>My param does things.</dd>
</dl>
</blockquote>

However, the defaults get lost when migrating to sphinx-autodoc-typehints. I know the defaults are in the function signature above the param docs, but for long signatures (and therefor long lists of params), it would be best to also have them in the param docs.

Ideally we would parse and use the source code for the defaults instead of going from python objects back to strings, but using `repr` to get OK representations for most defaults would also be OK.

flying-sheep:
WIP.

Fixes #48

### DIFF
--- a/sphinx_autodoc_typehints.py
+++ b/sphinx_autodoc_typehints.py
@@ -347,6 +347,7 @@ def process_docstring(app, what, name, obj, options, lines):
             obj = getattr(obj, '__init__')
 
         obj = inspect.unwrap(obj)
+        sig = inspect.signature(obj)
         type_hints = get_all_type_hints(obj, name)
 
         for argname, annotation in type_hints.items():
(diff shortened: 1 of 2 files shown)


gaborbernat:
This seems very outdated now, I'll close this as such, but we can reopen a new PR if we figure this one out.

flying-sheep:
I re-did the PR: #191

Would be nice if you could take a look this time.

PR #107, opened 2019-10-21
Introduce typehints_document_rtype config.

### LINKED ISSUE #106: Add option: Hide automatically inferred rtype
I use napoleon with the following settings:
```
napoleon_use_param = False
napoleon_use_keyword = False
napoleon_use_rtype = False
```

Therefore, there is no `:rtype:`, therefore sphinx-autodoc-typehints adds one. ("The :rtype: directive is added if and only if no existing :rtype: is found.")

I would like to have an option that disables this behavior, so I can use sphinx-autodoc-typehints
 solely to strip the annotations from the function signature.

moi90:
If typehints_document_rtype=False (default True), never add an :rtype: directive.

Closes #106.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -63,6 +63,8 @@ The following configuration options are accepted:
 * ``always_document_param_types`` (default: ``False``): If ``False``, do not add type info for
   undocumented parameters.  If ``True``, add stub documentation for undocumented parameters to
   be able to add type info.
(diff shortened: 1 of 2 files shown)


moi90:
You're welcome, thanks for merging!

agronholm:
I just released v1.9.0 and this feature is included.

PR #120, opened 2019-11-23
Add support for union args on 3.7

saulshanabrook:
On  3.7, the current union args will always be None, which raises an exception because `len(None)` is not allowed.

This changes it to look for `__args__` if its present.

Inspired by https://github.com/ilevkivskyi/typing_inspect/blob/45efedf7ea3c0debbed3a38ba0b43aefa346ff9a/typing_inspect.py#L335. I considered adding a dependency to typing_inspect, but this was previously ruled out https://github.com/agronholm/sphinx-autodoc-typehints/issues/82


agronholm:
I'm sorry, can you elaborate on what practical problem this PR solves?

agronholm:
Also a requirement would be a test that fails without the patch and passes with it.

saulshanabrook:
Sorry about this,  I realized it was an issue on my side monkeypatching getattr on `_GenericAlias`

PR #138, opened 2020-05-01
Remove use of sphinx's Signature class

### LINKED ISSUE #133: sphinx.util.inspect.Signature raises RemovedInSphinx40Warning
`sphinx.util.inspect.Signature` was deprecated in 2.4.0. Building docs in sphinx 3.0.0 with `sphinx-autodoc-typehints` therefore raises a ton of warnings of the following type:
```
RemovedInSphinx40Warning: sphinx.util.inspect.Signature() is deprecated
  warnings.warn('sphinx.util.inspect.Signature() is deprecated',
```

The dependency on `sphinx.util.inspect.Signature` should be removed, and instead the functionality from python's `inspect` module should be used.

mtreinish:
The sphinx.util.inspect.Signature api has been deprecated since sphinx
2.4 and is staged for removal in the 4.0 release. [1] This commit updates
the usage to use the alternative function for it that is not
deprecated (and returns a inspect.Signature object from stdlib). This should
avoid raising a deprecation warning on each function signature that is using
typehints.

Fixes #133

[1] https://www.sphinx-doc.org/en/latest/extdev/deprecated.html?highlight=signature

### DIFF
--- a/sphinx_autodoc_typehints.py
+++ b/sphinx_autodoc_typehints.py
@@ -5,7 +5,8 @@ import typing
 from typing import get_type_hints, TypeVar, Any, AnyStr, Tuple
 
 from sphinx.util import logging
-from sphinx.util.inspect import Signature
+from sphinx.util.inspect import signature as Signature
+from sphinx.util.inspect import stringify_signature
 
 logger = logging.getLogger(__name__)
 pydata_annotations = {'Any', 'AnyStr', 'Callable', 'ClassVar', 'Literal', 'NoReturn', 'Optional',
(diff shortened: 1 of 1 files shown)


agronholm:
You need to fix the test failures. Did the test suite pass locally for you?

agronholm:
Also, why doesn't this patch just directly import `inspect.signature`?

agronholm:
The test failures were not caused by this patch, but rather new semantics in Sphinx 3.0 so I've merged this. Thanks!

gaborbernat:
@agronholm can you please do a release with this? thanks!

PR #157, opened 2020-10-23
Support getting methods using metaclasses

flying-sheep:
If a metaclass has a `__getattr__` method, `__dict__` won’t cut it.

### DIFF
--- a/sphinx_autodoc_typehints.py
+++ b/sphinx_autodoc_typehints.py
@@ -253,7 +253,7 @@ def process_signature(app, what: str, name: str, obj, options, signature, return
                 class_name = obj.__qualname__.split('.')[-2]
                 method_name = "_{c}{m}".format(c=class_name, m=method_name)
 
-            method_object = outer.__dict__[method_name] if outer else obj
+            method_object = getattr(outer, method_name) if outer else obj
             if not isinstance(method_object, (classmethod, staticmethod)):
                 del parameters[0]
 


agronholm:
The test suite fails, plus you have not added any tests of your own. Fix those problems and I will review this.

flying-sheep:
Hmm, seems like I can’t:

Both static methods and normal methods return `types.FunctionType` instances on `getattr` (class methods return `types.MethodType` objects aka bound methods.)

We could use `object.getattribute(outer, method_name)`, but that one probably just uses `__dict__` internally…

gaborbernat:
@flying-sheep  I've rebased on main, can you please fix the CI, thanks!

flying-sheep:
You didn’t!

gaborbernat:
I did but then further changes were made to main that made it out of date.

flying-sheep:
Interesting that such a trivial change can have conflicts at all. I just remade it.

gaborbernat (changes requested):
Can we have a test and a changelog for this, please?

gaborbernat:
@flying-sheep can you please force push back 9ace51eade7dda1e9b98b6266b4316bf6cc4b7be, I by mistake overwritten this branch. Thanks :pray:

gaborbernat:
On a deeper look, I'm actually not sure how this can happen. The method must already be manifested because otherwise it would not be passed in. I'm closing until we can come up with a case that replicates it :thinking:

PR #165, opened 2021-03-27
Bugfix for Python 3.9

zwang123:
## Overview

In Python 3.9, `typing.Callable` is deprecated, in favor of `collections.abc.Callable`. Thus, the code needs to be updated accordingly.

## Reason

`get_type_hints` for `collections.abc.Callable` returns in format `[[…], …]` instead of `[…, …]`, the latter of which is the case for `typing.Callable`.

Additionally, `collections.abc.Callable` is a `class` instead of `data`.

## Example

With `typehints_fully_qualified = True` for clarity.

Tested on `Python 3.9.1 (tags/v3.9.1:1e5d33e, Dec  7 2020, 17:08:21) [MSC v.1927 64 bit (AMD64)] on win32`.

### DIFF
--- a/sphinx_autodoc_typehints.py
+++ b/sphinx_autodoc_typehints.py
@@ -143,6 +143,14 @@ def format_annotation(annotation,
             for arg in args[:-1]) + ']'
         formatted_args += ', ' + format_annotation(
             args[-1], simplify_optional_unions=simplify_optional_unions) + ']'
+    elif full_name == 'collections.abc.Callable' and args and args[0] is not ...:
+        assert len(args) == 2
+        formatted_args = '\\[\\[' + ', '.join(
+            format_annotation(
+                arg, simplify_optional_unions=simplify_optional_unions)
+            for arg in args[0]) + ']'
+        formatted_args += ', ' + format_annotation(
+            args[-1], simplify_optional_unions=simplify_optional_unions) + ']'
     elif full_name == 'typing.Literal':
         formatted_args = '\\[' + ', '.join(repr(arg) for arg in args) + ']'
 
--- a/sphinx_autodoc_typehints.py
+++ b/sphinx_autodoc_typehints.py
@@ -118,7 +118,7 @@ def format_annotation(annotation,
 
     full_name = (module + '.' + class_name) if module != 'builtins' else class_name
     prefix = '' if fully_qualified or full_name == class_name else '~'
-    role = 'data' if class_name in pydata_annotations else 'class'
+    role = 'data' if class_name in pydata_annotations and module != 'collections.abc' else 'class'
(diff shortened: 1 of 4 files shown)


agronholm:
Please add a test that fails without this fix.

agronholm:
Also please rebase against master, as the previously failing tests have been fixed there now.

zwang123:
I have added the test case here. But the check failed.

I tried to test on the current master branch, but got the same warning. I suppose this is not due to my changes.

agronholm:
The spot where the check failed relates directly to your patch. As for the warnings you see, those are normal.

agronholm:
You should probably test with `tox` locally to see if the tests pass before pushing more commits to the PR.

zwang123:
> You should probably test with `tox` locally to see if the tests pass before pushing more commits to the PR.

Thanks for the comments. I will do it next time.

agronholm:
I checked and the tests pass even without this patch.

agronholm:
Never mind, I didn't patch all the necessary files.

sphinx_autodoc_typehints.py
```diff
@@ -129,11 +129,13 @@ def format_annotation(annotation,
...
             args = tuple(x for x in args if x is not type(None))  # noqa: E721
-    elif full_name == 'typing.Callable' and args and args[0] is not ...:
+    elif full_name in ['typing.Callable', 'collections.abc.Callable'] \
+            and args and args[0] is not ...:
+        arglist = args[0] if isinstance(args[0], list) else args[:-1]
         formatted_args = '\\[\\[' + ', '.join(
             format_annotation(
                 arg, simplify_optional_unions=simplify_optional_unions)
-            for arg in args[:-1]) + ']'
+            for arg in arglist) + ']'
```
agronholm, line 138: These changes seem to be unnecessary. The first element of `args` is never a list. I reverted this and your new tests still pass.

tests/roots/test-dummy/dummy_module.py
```diff
@@ -1,7 +1,13 @@
+import sys
 import typing
 from dataclasses import dataclass
 from mailbox import Mailbox
-from typing import Callable, Union
+from typing import Any, Callable, Union
+
+if sys.version_info.major == 3 and sys.version_info.minor >= 9:
```
agronholm, line 7:
```suggestion
if sys.version_info >= (3, 9):
```

gaborbernat:
@zwang123  I've rebased on main, can you please fix the CI, thanks!

gaborbernat:
Closing due to inactivity, please reopen a new PR in case you'd like to pick it up again. Thanks for the understanding!

Issue #129, opened 2020-02-18
Unexpected indent


ppwwyyxx:
I received a stack trace like this when using this project:
```
  File "/home/XXX/.local/lib/python3.8/site-packages/sphinx/ext/autodoc/__init__.py", line 731, in generate
    self.add_content(more_content)
  File "/home/XXX/.local/lib/python3.8/site-packages/sphinx/ext/autodoc/__init__.py", line 470, in add_content
    for i, line in enumerate(self.process_doc(docstrings)):
  File "/home/XXX/.local/lib/python3.8/site-packages/sphinx/ext/autodoc/__init__.py", line 438, in process_doc
    self.env.app.emit('autodoc-process-docstring',
  File "/home/XXX/.local/lib/python3.8/site-packages/sphinx/application.py", line 447, in emit
    return self.events.emit(event, *args)
  File "/home/XXX/.local/lib/python3.8/site-packages/sphinx/events.py", line 99, in emit
    results.append(callback(self.app, *args))
  File "/home/XXX/.local/lib/python3.8/site-packages/sphinx_autodoc_typehints.py", line 362, in process_docstring
    type_hints = get_all_type_hints(obj, name)
  File "/home/XXX/.local/lib/python3.8/site-packages/sphinx_autodoc_typehints.py", line 217, in get_all_type_hints
    rv = backfill_type_hints(obj, name)
  File "/home/XXX/.local/lib/python3.8/site-packages/sphinx_autodoc_typehints.py", line 258, in backfill_type_hints
    obj_ast = ast.parse(textwrap.dedent(inspect.getsource(obj)), **parse_kwargs)
  File "/usr/lib/python3.8/ast.py", line 47, in parse
    return compile(source, filename, mode, flags,
  File "<unknown>", line 1
    @staticmethod
    ^
IndentationError: unexpected indent

```

I think the issue is due to the use of `textwrap.dedent` to dedent code. The function that causes the error look like this:
```
    @staticmethod
    def get(name):
        """
        Args:
            name (str): name of a dataset (e.g. coco_2014_train).

        Returns:
            Metadata: The :class:`Metadata` instance associated with this name,
            or create an empty one if none is available.
        """
        assert len(name)
        if name in MetadataCatalog._NAME_TO_META:
            ret = MetadataCatalog._NAME_TO_META[name]
            # TODO this is for the BC breaking change in D15247032.
            # Remove this in the future.
            if hasattr(ret, "dataset_name"):
[...]

PR #166, opened 2021-03-27
fix indent of multiline string

ppwwyyxx:
close #127 and #129

### DIFF
--- a/sphinx_autodoc_typehints.py
+++ b/sphinx_autodoc_typehints.py
@@ -159,6 +159,45 @@ def format_annotation(annotation,
         role=role, prefix=prefix, full_name=full_name, formatted_args=formatted_args)
 
 
+# reference: https://github.com/pytorch/pytorch/pull/46548/files
+def normalize_source_lines(sourcelines: str) -> str:
+    """
+    This helper function accepts a list of source lines. It finds the
+    indentation level of the function definition (`def`), then it indents
+    all lines in the function body to a point at or greater than that
+    level. This allows for comments and continued string literals that
+    are at a lower indentation than the rest of the code.
+    Arguments:
+        sourcelines: source code
+    Returns:
+        source lines that have been correctly aligned
+    """
+    sourcelines = sourcelines.split("\n")
+
+    def remove_prefix(text, prefix):
+        return text[text.startswith(prefix) and len(prefix):]
+
+    # Find the line and line number containing the function definition
+    for i, l in enumerate(sourcelines):
+        if l.lstrip().startswith("def"):
+            idx = i
+            break
+    else:
+        return "\n".join(sourcelines)
+    fn_def = sourcelines[idx]
+
+    # Get a string representing the amount of leading whitespace
+    whitespace = fn_def.split("def")[0]
+
+    # Add this leading whitespace to all lines before and after the `def`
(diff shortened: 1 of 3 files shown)


agronholm:
I'll need a least one test that fails without this fix.

ppwwyyxx:
Tets was added

agronholm:
Thank you!

kylepjohnson:
Thank you @ppwwyyxx !

Do you know when you will make the next release to PyPI?

agronholm:
Sorry for taking so long but it's out now.

shoyer:
I'm not quite sure why, but it looks like this change broke JAX's doc build: https://github.com/google/jax/pull/6449

shoyer:
For reference, here is the error message that turns up in our logs:
```
Traceback (most recent call last):
  File "/home/docs/checkouts/readthedocs.org/user_builds/jax/envs/latest/lib/python3.8/site-packages/sphinx/events.py", line 111, in emit
    results.append(listener.handler(self.app, *args))
  File "/home/docs/checkouts/readthedocs.org/user_builds/jax/envs/latest/lib/python3.8/site-packages/sphinx_autodoc_typehints.py", line 417, in process_docstring
    type_hints = get_all_type_hints(obj, name)
  File "/home/docs/checkouts/readthedocs.org/user_builds/jax/envs/latest/lib/python3.8/site-packages/sphinx_autodoc_typehints.py", line 271, in get_all_type_hints
    rv = backfill_type_hints(obj, name)
  File "/home/docs/checkouts/readthedocs.org/user_builds/jax/envs/latest/lib/python3.8/site-packages/sphinx_autodoc_typehints.py", line 312, in backfill_type_hints
    obj_ast = ast.parse(textwrap.dedent(
  File "/home/docs/.pyenv/versions/3.8.6/lib/python3.8/ast.py", line 47, in parse
    return compile(source, filename, mode, flags,
  File "<unknown>", line 12
    max_rows_cols = max(a.shape[-2:])
                                    ^
IndentationError: expected an indented block
```

And the apparent source line:
https://github.com/google/jax/blob/32564442cbab5bec712cadc316a1c01396faf89c/jax/_src/numpy/linalg.py#L304

ppwwyyxx:
Because the decorator contains:
```
    default `rcond` is `1e-15`. Here the default is
```
that starts with "def", the code wrongly assumes this is the function definition line.

Not sure what is a more robust way to find the function definition line of a function source code.

shoyer:
I would look into Python AST, or maybe whatever approaches are used by syntax highlighters. But certainly a line starting with "def" is not a reliable indication :)

ppwwyyxx:
The point of this PR is because we can't use python ast: ast fails to parse a string like
```
    def my_method(self):
        a = '''long
string with bad indentation'''
        return a
```
because the string has to be dedented first. To dedent the string reliably we need to know what's the correct indentation of the function def line.

shoyer:
So I'm not sure how, but both Google Colab and VSCode seem correctly infer the syntax highlighting here:
image

shoyer:
I wonder if the simpler approach of just taking the indentation from the first line in the source code would work. If I understand it correctly, that is always either the function definition itself or a decorator, which must also be intended the same amount as the function definition.

agronholm:
Should I yank the release from PyPI?

PR #171, opened 2021-04-17
Configurable formatter

### LINKED ISSUE #112: Allow customizing formatter
Before making a PR, I’d like to check if you’d be open to the idea.

Basically, instead of `format_annotation(annotation, fully_qualified=False)`, we’d have

```py
class AnnotationFormatter:
    def __init__(self, *, fully_qualified=False):
        self.fully_qualified = fully_qualified
    def format(self, annotation) -> str:
        ...
```

Also a new config option `typehints_format_cls` where people would have the ability to override formatting:

```py
class CustomFormatter(sat.AnnotationFormatter):
    def format(self, annotation) -> str:
        if isinstance(annotation, MyClass):
            return ...
        return super().format(annotation)

typehints_format_cls = CustomFormatter
```

For backwards compatibility (if desirable), we could define

```py
def format_annotation(annotation, fully_qualified=False):
    return AnnotationFormatter(fully_qualified=fully_qualified).format(annotation)
```

flying-sheep:
Fixes #112


gaborbernat:
@flying-sheep I've rebased on main, can you please fix the CI, thanks!

gaborbernat (changes requested):
The CI is failing.

gaborbernat (changes requested):
Requires test, changelog and documentation.

flying-sheep:
I’ll wait for the test module in #191 to land, that’ll make testing easier.

Can you please help with the tests there?

src/sphinx_autodoc_typehints/__init__.py
```diff
@@ -97,7 +98,17 @@ def get_annotation_args(annotation: Any, module: str, class_name: str) -> tuple[
...
-def format_annotation(annotation: Any, fully_qualified: bool = False, simplify_optional_unions: bool = True) -> str:
+def format_annotation(
+    annotation: Any,
+    fully_qualified: bool = False,
+    simplify_optional_unions: bool = True,
+    typehints_formatter: Callable[..., str] | None = None,
+) -> str:
+    # all options except for the formatter:
+    opts = {"fully_qualified": fully_qualified, "simplify_optional_unions": simplify_optional_unions}
+    formatted = None if typehints_formatter is None else typehints_formatter(annotation, **opts)
```
gaborbernat, line 109: Let's not build the opts dictionary but instead pass it in explicitly here as keyword arguments.

src/sphinx_autodoc_typehints/__init__.py
```diff
@@ -143,14 +154,16 @@ def format_annotation(annotation: Any, fully_qualified: bool = False, simplify_o
             args_format = f"\\[:py:data:`{prefix}typing.Union`\\[{{}}]]"
             args = tuple(x for x in args if x is not type(None))  # noqa: E721
     elif full_name == "typing.Callable" and args and args[0] is not ...:
-        fmt = ", ".join(format_annotation(arg, simplify_optional_unions=simplify_optional_unions) for arg in args[:-1])
+        opts_shortened = dict(opts, typehints_formatter=typehints_formatter)
+        del opts_shortened["fully_qualified"]
+        fmt = ", ".join(format_annotation(arg, **opts_shortened) for arg in args[:-1])  # type: ignore # arg spec
         formatted_args = f"\\[\\[{fmt}]"
-        formatted_args += f", {format_annotation(args[-1], simplify_optional_unions=simplify_optional_unions)}]"
+        formatted_args += f", {format_annotation(args[-1], **opts_shortened)}]"  # type: ignore # arg spec
```
gaborbernat, line 161: revert this

src/sphinx_autodoc_typehints/__init__.py
```diff
@@ -143,14 +154,16 @@ def format_annotation(annotation: Any, fully_qualified: bool = False, simplify_o
...
     elif full_name == "typing.Literal":
         formatted_args = f"\\[{', '.join(repr(arg) for arg in args)}]"
 
     if args and not formatted_args:
-        fmt = ", ".join(format_annotation(arg, fully_qualified, simplify_optional_unions) for arg in args)
+        fmt = ", ".join(format_annotation(arg, **opts, typehints_formatter=typehints_formatter) for arg in args)
```
gaborbernat, line 166: revert this

src/sphinx_autodoc_typehints/__init__.py
```diff
@@ -528,6 +538,10 @@ def validate_config(app: Sphinx, env: BuildEnvironment, docnames: list[str]) ->
     if app.config.typehints_defaults not in valid | {False}:
         raise ValueError(f"typehints_defaults needs to be one of {valid!r}, not {app.config.typehints_defaults!r}")
 
+    formatter = app.config.typehints_formatter
+    if formatter is not None and not callable(formatter):
+        raise ValueError(f"typehints_formatter needs to be callable or `None`, not {formatter}")
```
gaborbernat, line 543: Needs test.

gaborbernat (changes requested):

src/sphinx_autodoc_typehints/__init__.py
flying-sheep: OK, but to keep the lines short, I made it a partial in the `typing.Callable` branch

src/sphinx_autodoc_typehints/__init__.py
flying-sheep: already has one!

https://github.com/flying-sheep/sphinx-autodoc-typehints/blob/6601ef0029f6de46c0f7c1c0c573cdb51cb1cb52/tests/test_sphinx_autodoc_typehints.py#L673

gaborbernat:
@flying-sheep can you please force push back 3e6a36c5f002ed98ebf7708f37f0700c3018d081, I by mistake overwritten this branch. Thanks :pray:

flying-sheep:
All done!

src/sphinx_autodoc_typehints/__init__.py
```diff
@@ -89,7 +89,19 @@ def get_annotation_args(annotation: Any, module: str, class_name: str) -> tuple[
...
-def format_annotation(annotation: Any, fully_qualified: bool = False, simplify_optional_unions: bool = True) -> str:
+def format_annotation(
+    annotation: Any,
+    fully_qualified: bool = False,
+    simplify_optional_unions: bool = True,
+    typehints_formatter: Callable[..., str] | None = None,
+) -> str:
+    if typehints_formatter is not None:
+        formatted = typehints_formatter(
+            annotation, fully_qualified=fully_qualified, simplify_optional_unions=simplify_optional_unions
```
gaborbernat, line 100: Finish this with a comma to force split into new lines (to be consistent with other instances)

src/sphinx_autodoc_typehints/__init__.py
```diff
@@ -89,7 +89,21 @@ def get_annotation_args(annotation: Any, module: str, class_name: str) -> tuple[
...
+    annotation: Any,
+    fully_qualified: bool = False,
+    simplify_optional_unions: bool = True,
+    typehints_formatter: Callable[..., str] | None = None,
+) -> str:
+    if typehints_formatter is not None:
+        formatted = typehints_formatter(
+            annotation,
+            fully_qualified=fully_qualified,
+            simplify_optional_unions=simplify_optional_unions,
```
gaborbernat, line 102: On second thought let's pass in the `app.config` argument here as is. That way we'll not have to keep adding arguments in the future.

gaborbernat (changes requested):

PR #189, opened 2022-01-01
Drop Python 3.7 and adopt new best practices

gaborbernat:
Signed-off-by: Bernát Gábor <gaborjbernat@gmail.com>


agronholm:
I think this was supposed to be named "Drop Python 3.6..."

gaborbernat:
Oh, 🤦‍♂️ indeed sorry

PR #191, opened 2022-01-04
Add defaults

### LINKED ISSUE #48: Also annotate parameter defaults.

flying-sheep:
Supersedes #103, fixes #48

In the previous PR, I added them to types, this time I’m adding it to the string. I think both have arguments pro and contra. Should it be configurable?

### DIFF
--- a/src/sphinx_autodoc_typehints/__init__.py
+++ b/src/sphinx_autodoc_typehints/__init__.py
@@ -427,8 +427,6 @@ def split_type_comment_args(comment):
 
 
 def format_default(app: Sphinx, default: Any) -> Optional[str]:
-    if not app.config.typehints_defaults:
-        return None
     if default is inspect.Parameter.empty:
         return None
     formatted = repr(default).replace("\\", "\\\\")
--- a/src/sphinx_autodoc_typehints/__init__.py
+++ b/src/sphinx_autodoc_typehints/__init__.py
@@ -2,7 +2,7 @@ import inspect
 import sys
 import textwrap
 import typing
-from typing import Any, AnyStr, NewType, Tuple, TypeVar, get_type_hints
+from typing import Any, AnyStr, Dict, NewType, Optional, Tuple, TypeVar, get_type_hints
 
 from sphinx.application import Sphinx
 from sphinx.util import logging
@@ -426,7 +426,7 @@ def split_type_comment_args(comment):
     return result
 
 
-def format_default(app: Sphinx, default):
+def format_default(app: Sphinx, default: Any) -> Optional[str]:
     if not app.config.typehints_defaults:
         return None
     if default is inspect.Parameter.empty:
@@ -516,7 +516,7 @@ def process_docstring(app: Sphinx, what, name, obj, options, lines):  # noqa: U1
                 lines.insert(insert_index, f":rtype: {formatted_annotation}")
 
 
-def builder_ready(app: Sphinx):
+def builder_ready(app: Sphinx) -> None:
     if app.config.set_type_checking_flag:
         typing.TYPE_CHECKING = True
     valid = {None, "comma", "braces", "braces-after"}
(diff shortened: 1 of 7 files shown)


gaborbernat:
I'd like to have it by default off. Otherwise, I'm happy to also allow it to be configurable where it's added.

flying-sheep:
OK, I need help, I have no idea what’s going on with those `"`s and `*`s.

gaborbernat:
> OK, I need help, I have no idea what’s going on with those `"`s and `*`s.
> 
> Why does only my test case fail on Python 3.9 even though I copied the setup from `test_sphinx_output_future_annotations`?

That's half the work in this project, to figure out the differences between Python versions and make it work everywhere. I'll have a look at some point, but no idea when I'll have time for this.

flying-sheep:
Ugh, I figured it out. This will bite you in the future! So even on 3.9, a simple type like `int` gets compiled to

- `*int*` if the file *contains* new-style union syntax, and
- `"int"` otherwise.

It’s *not* about `from __future__ import annotations`, it’s about actually using the syntax! So:

```py
from __future__ import annotations

def function(x: bool) -> str | int: ...
```

gets compiled to `*`

```
...
Parameters:
   * **x** (*bool*) -- foo
```

And when you don‘t use `|` syntax:

```py
from __future__ import annotations

def function(x: bool) -> str: ...
```

gets compiled to `"`

```
...
Parameters:
   * **x** ("bool") -- foo
```

gaborbernat:
> Ugh, I figured it out. This will bite you in the future, I hope you find a way to make it less flaky.
> 
> I filed #194 to explore this. But this PR is now ready!

Sorry, I'm not open to a fallback issue and working around the test. We should figure out here and now why this is the case and avoid it.

flying-sheep:
I’m avoiding it by not using that syntax, in the same way your main test does. This PR is completely independent of this problem now, I just happened to run into it while developing this PR.

src/sphinx_autodoc_typehints/__init__.py
```diff
@@ -493,16 +516,20 @@ def process_docstring(app, what, name, obj, options, lines):  # noqa: U100
                 lines.insert(insert_index, f":rtype: {formatted_annotation}")
 
 
-def builder_ready(app):
+def builder_ready(app: Sphinx):
```
gaborbernat, line 519: Let's type-hint the return value to if you're going to type annotate it.

src/sphinx_autodoc_typehints/__init__.py
```diff
@@ -493,16 +516,20 @@ def process_docstring(app, what, name, obj, options, lines):  # noqa: U100
...
     if app.config.set_type_checking_flag:
         typing.TYPE_CHECKING = True
+    valid = {None, "comma", "braces", "braces-after"}
+    if app.config.typehints_defaults not in valid | {False}:
+        raise ValueError(f"typehints_defaults needs to be one of {valid!r}, not {app.config.typehints_defaults!r}")
```
gaborbernat, line 524: test for this please

src/sphinx_autodoc_typehints/__init__.py
```diff
@@ -425,7 +426,19 @@ def add(val):
     return result
 
 
-def process_docstring(app, what, name, obj, options, lines):  # noqa: U100
+def format_default(app: Sphinx, default):
+    if not app.config.typehints_defaults:
+        return None
```
gaborbernat, line 431: If this is the case the method will never be called so I don't think we need this

gaborbernat (changes requested):

Issue #139, opened 2020-05-09
Getting an `IndentationError` related specifically to a doc string while building.


demitri:
I've encountered a Sphinx bug which I reported [here](https://github.com/sphinx-doc/sphinx/issues/7631), and it was pointed out to me (via the log) that the error was raised by sphinx-autodoc-hints. The report is duplicated below.

I'm getting the following error when trying to build HTML files:

```
/usr/local/anaconda/lib/python3.7/site-packages/sphinx/util/inspect.py:555: RemovedInSphinx40Warning: sphinx.util.inspect.Signature() is deprecated
  RemovedInSphinx40Warning)

Exception occurred:
  File "/usr/local/anaconda/lib/python3.7/site-packages/typed_ast/ast3.py", line 62, in parse
    return _ast3._parse(source, filename, mode, feature_version)
  File "<unknown>", line 1
    @classmethod
    ^
IndentationError: unexpected indent
The full traceback has been saved in /var/folders/yp/lf1xdvp165n5xb7hcrwzpngc0000gn/T/sphinx-err-jidws8_h.log, if you want to report the issue to the developers.
Please also report this if it was a user error, so that a better error message can be provided next time.
A bug report can be filed in the tracker at <https://github.com/sphinx-doc/sphinx/issues>. Thanks!
make: *** [html] Error 2
```

The code that it's getting caught on is this:

```python
	@classmethod
	def fromFITSHeader(cls, fits_header=None, uncertainty=4.848e-6):
		'''
		Factory method to create a region from the provided FITS header; the returned object will be as specific as possible (but probably an ASTPolygon).
		
		:param fits_header: a FITS header (Astropy, fitsio, an array of cards)
		:param uncertainty: defaults to 4.848e-6, an uncertainty of 1 arcsec
		'''
		if fits_header is None:
			raise ValueError("This method requires a 'fits_header' to be set.")

		. . .
```

I am sure the problem is not an indentation error (at least as far as the Python code). When I delete the entire docstring, `make html` works without error. *Any* docstring I add causes the error, even a blank one, e.g.

```
'''

'''
```

The Sphinx log is attached here:
[sphinx-err-gnkuj5m5.log](https://github.com/sphinx-doc/sphinx/files/4596009/sphinx-err-gnkuj5m5.log)

Thanks!

amarandon:
[...]

PR #198, opened 2022-01-06
Fixed normalize_source_lines messing with the indentation of method

scganterh:
Fixed normalize_source_lines messing with the indentation of method with decorator that has parameter starting with `def`.

`normalize_source_lines` was breaking a method like the following:

```python
    @_with_parameters(
        _Parameter("self", _Parameter.POSITIONAL_OR_KEYWORD),
        *_proxy_instantiation_parameters,
        _project_id,
        _Parameter(
            "node_numbers",
            _Parameter.POSITIONAL_OR_KEYWORD,
            default=None,
            annotation=Optional[Iterable[int]],
        ),
    )
    def __init__(bound_args):  # noqa: N805
        ...
```

Note the decorator parameter `default=None,` starts with `def` and confuses the function.

I got to this fix from #139.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Changelog
 
+## dev
+
+- Fixed `normalize_source_lines()` messing with the indentation of methods with decorators that have parameters starting with `def`.
+
 ## 1.14.0
 
 - Added `typehints_defaults` config option allowing to automatically annotate parameter defaults.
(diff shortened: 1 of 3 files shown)


gaborbernat (changes requested):
Needs a test and changelog entry.

scganterh:
> Needs a test and changelog entry.

I am not sure how to add the changelog entry. Should I add a new version?

gaborbernat:
> I am not sure how to add the changelog entry. Should I add a new version?

 Just use version name dev 👍

gaborbernat (approved):

PR #200, opened 2022-01-07
Fix KeyError with type hints on the first arg of bound methods

aiudirog:
https://github.com/tox-dev/sphinx-autodoc-typehints/commit/8dc8d94912431fa79a13f39d7570637244ac1a01 introduced an issue where a bound method with a type hint on the first argument would raise a KeyError looking up the default because that argument isn't present in the signature. This PR simply ignores those parameters.


aiudirog:
Thanks for the fast turnaround!

PR #210, opened 2022-01-14
inspect.getsource can raise TypeError

### LINKED ISSUE #209: Uncaught exception for C++ module
Although the documentation of `inspect.getsource` says
> An OSError is raised if the source code cannot be retrieved.

When I do
```
import inspect, sys
inspect.getsource(sys)
```
I get a TypeError and not a OSError, thus generating the documentation with a C++ module fails (it did not for earlier versions).

Catching both TypeError and OSError here fixes the issue:
https://github.com/tox-dev/sphinx-autodoc-typehints/blob/01b3a1abeac063a5530b96b78252aa17a93f040c/src/sphinx_autodoc_typehints/__init__.py#L286

jmirabel:
Fixes #209

### DIFF
--- a/src/sphinx_autodoc_typehints/__init__.py
+++ b/src/sphinx_autodoc_typehints/__init__.py
@@ -283,7 +283,7 @@ def _resolve_type_guarded_imports(obj: Any) -> None:
             if module:
                 try:
                     module_code = inspect.getsource(module)
-                except OSError:
+                except (TypeError, OSError):
                     ...  # no source code => no type guards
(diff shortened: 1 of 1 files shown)


gaborbernat (changes requested):
Waiting for test and reproducible.

jmirabel:
Thank you.

gaborbernat:
https://github.com/tox-dev/sphinx-autodoc-typehints/releases/tag/1.17.0

PR #211, opened 2022-01-18
Prevents reaching inner blocks that contains `if TYPE_CHECKING`

thepabloaguilar:
Hi, thanks for your work on this plugin! We're having some problems with [returns](https://github.com/dry-python/returns) because we use type guards inside some classes like [here](https://github.com/dry-python/returns/blob/master/returns/primitives/hkt.py#L92) which leads in some warnings while building sphinx broking our builds:

```text
WARNING: Failed guarded type import with AttributeError('__globals__')
```

This PR fixes that issue, basically when a class object doesn't have the `__globals__` attribute so we basically can call `exec` without the globals if the given object doesn't have the mentioned attribute.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Changelog
 
+## WIP
+
+- Prevents reaching inner blocks that contains `if TYPE_CHECKING`
+
 ## 1.15.2
 
 - Log a warning instead of crashing when a type guard import fails to resolve
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,6 @@
 # Changelog
 
-## WIP
+## 1.15.3
 
 - Prevents reaching inner blocks that contains `if TYPE_CHECKING`
 
--- a/src/sphinx_autodoc_typehints/__init__.py
+++ b/src/sphinx_autodoc_typehints/__init__.py
@@ -271,7 +271,7 @@ def get_all_type_hints(obj: Any, name: str) -> dict[str, Any]:
     return _get_type_hint(name, obj)
 
 
-_TYPE_GUARD_IMPORT_RE = re.compile(r"if (typing.)?TYPE_CHECKING:[^\n]*([\s\S]*?)(?=\n\S)")
+_TYPE_GUARD_IMPORT_RE = re.compile(r"\nif (typing.)?TYPE_CHECKING:[^\n]*([\s\S]*?)(?=\n\S)")
 _TYPE_GUARD_IMPORTS_RESOLVED = set()
 
 
--- a/tests/roots/test-resolve-typing-guard/demo_typing_guard.py
+++ b/tests/roots/test-resolve-typing-guard/demo_typing_guard.py
@@ -29,6 +29,15 @@ def a(f: Decimal, s: AnyStr) -> Sequence[AnyStr | Decimal]:
     return [f, s]
 
 
+class SomeClass:
+    """This class do something."""
+
+    if TYPE_CHECKING:  # Classes doesn't have `__globals__` attribute
+
+        def __getattr__(self, item: str):  # noqa: U100
+            """This method do something."""
+
+
 __all__ = [
     "a",
     "ValueError",


tests/roots/test-resolve-typing-guard/demo_typing_guard.py
```diff
@@ -29,6 +29,15 @@ def a(f: Decimal, s: AnyStr) -> Sequence[AnyStr | Decimal]:
     return [f, s]
 
 
+class SomeClass:
+    """This class do something."""
+
+    if TYPE_CHECKING:  # Classes doesn't have `__globals__` attribute
+
+        def __getattr__(self, item: str):  # noqa: U100
```
gaborbernat, line 37: This is an interesting example. However, I don't think we should be evaluating these blocks; our evaluation scope is global and this is clearly tied to a class. Perhaps we should ensure that  `if TYPE_CHECKING` is not preceded by any whitespace to avoid such scooping issues.

gaborbernat (changes requested):
Can you add a changelog entry too, thanks!

tests/roots/test-resolve-typing-guard/demo_typing_guard.py
gaborbernat: One downside here would be that we'd not be able to generate type information for methods defined as such. I don't think though we have a better solution here as often time the reason for using `TYPE_CHECKING` is to avoid circular dependencies, so if we'd set this to `True` we'd fail to import the module altogether. For the time being I'd consider such code unsupported by the plugin 🤔 and perhaps we should just ignore them to not generate the warning.

thepabloaguilar: But if we ignore that code we'll get more errors, `if TYPE_CHECKING` is more than avoiding circular import. It's about adapting the code to be checked correctly in the type-checking phase. It's also used with `if not TYPE_CHECKING`.

gaborbernat: > But if we ignore that code we'll get more errors,

Such as?

> It's also used with `if not TYPE_CHECKING`.

Haven't seen this until now, and I'd say if you're using it like that you're doing something unsupported 🤔 

We cannot evaluate such nested (method/class) blocks because we don't have access to the scope where to populate it (only for global we have easy access to know the scope). And we cannot just set type_checking to True because that might break due to circular imports. Perhaps we should have a flag to switch between:

- set type checking true for a file and don't support circular import cases
- evaluate global scope as follow-up (what we do now) and do not support nested type check flags - handles circular import problems

thepabloaguilar: > I'd say if you're using it like that you're doing something unsupported 

Yeah, that was what I meant. If `mypy`, for example, doesn't support some advanced features and we need to prevent the code to be executed in the type-checking phase!
I can change those modifications to:

```python
# CODE
if hasattr(obj, "__globals__"):
    exec(guarded_code, obj.__globals__)
# CODE
```

This way we prevent any non-module code to be executed!

gaborbernat: I'd still prefer to alter the regex, I'd want to not even get to this state for such code.

gaborbernat: yeah 🤔

thepabloaguilar: Done, I've made the tests and adding just a `\n` at the begging do the work!

gaborbernat (changes requested):

src/sphinx_autodoc_typehints/__init__.py
```diff
@@ -271,7 +271,7 @@ def get_all_type_hints(obj: Any, name: str) -> dict[str, Any]:
     return _get_type_hint(name, obj)
 
 
-_TYPE_GUARD_IMPORT_RE = re.compile(r"if (typing.)?TYPE_CHECKING:[^\n]*([\s\S]*?)(?=\n\S)")
```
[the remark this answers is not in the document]
thepabloaguilar: How a file will start with a type guard without importing `TYPE_CHECKING`?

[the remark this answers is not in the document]
gaborbernat: You're right 🤦

gaborbernat (changes requested):
Let's just add a changelog entry and we're good 👍

CHANGELOG.md
```diff
@@ -1,5 +1,9 @@
 # Changelog
 
+## WIP
```
gaborbernat, line 3: 1.15.2

gaborbernat (approved):

gaborbernat:
https://github.com/tox-dev/sphinx-autodoc-typehints/actions/runs/1714063674

PR #212, opened 2022-01-20
Add handling of tuples in type subscriptions

### LINKED ISSUE #162: Sphinx cross-references not substituted at deeper levels of nesting for types with nested iterables
I'm looking to use `sphinx-autodoc-typehints` with `nptyping`, and it seems that for types with nested iterables (which will be every use of the `NDArray` type), while the types are correctly resolved out of the annotation syntax at all levels of nesting, the Sphinx cross-reference substitutions are only made at the outermost level of nesting:

For example, with a signature and docstring of:

```
def basis_func(cls, eta: NDArray[(Any, ...), Float], *, n: int) -> NDArray[(Any, ...), Float]:
    """Generate the basis function over `eta` at the given `n`.

    :param eta: Array (multi-dimensional ok) of spatial position values
        over which to calculate basis function values

    :param n: Index value to use in basis function calculation

    """
```

The generated output is:

image

----

This is with Python 3.9.1 and:

```
Sphinx==3.4.3
sphinx-autodoc-typehints==1.11.1
nptyping==1.4.0
```

bskinn:
Fixes #162.

Initial implementation and tests added, and project lints fixed.

There are some significant questions to be resolved before merging. Prime among them is whether introducing `nptyping` as a test/dev dependency is a good idea.

Also, input on whether this implementation is sufficiently general, and whether additional/different test cases should be added/used would be good.

### DIFF
--- a/src/sphinx_autodoc_typehints/__init__.py
+++ b/src/sphinx_autodoc_typehints/__init__.py
@@ -89,7 +89,7 @@ def get_annotation_args(annotation: Any, module: str, class_name: str) -> tuple[
     return getattr(annotation, "__args__", ())
 
 
-def format_internal_tuple(t: tuple, config: Config) -> str:
+def format_internal_tuple(t: tuple[Any, ...], config: Config) -> str:
     # An annotation can be a tuple, e.g., for nptyping:
     #   NDArray[(typing.Any, ...), Float]
     # In this case, format_annotation receives:
--- a/src/sphinx_autodoc_typehints/__init__.py
+++ b/src/sphinx_autodoc_typehints/__init__.py
@@ -94,8 +94,7 @@ def format_internal_tuple(t: tuple, config: Config) -> str:
     #   NDArray[(typing.Any, ...), Float]
     # In this case, format_annotation receives:
     #   (typing.Any, Ellipsis)
-    # This solution should hopefully be general for *any* type that allows
-    # tuples in annotations
+    # This solution should hopefully be general for *any* type that allows tuples in annotations
     fmt = [format_annotation(a, config) for a in t]
     if len(fmt) == 0:
         return "()"
--- a/src/sphinx_autodoc_typehints/__init__.py
+++ b/src/sphinx_autodoc_typehints/__init__.py
@@ -89,6 +89,22 @@ def get_annotation_args(annotation: Any, module: str, class_name: str) -> tuple[
     return getattr(annotation, "__args__", ())
 
 
+def format_internal_tuple(t: tuple, config: Config) -> str:
+    # An annotation can be a tuple, e.g., for nptyping:
+    #   NDArray[(typing.Any, ...), Float]
+    # In this case, format_annotation receives:
+    #   (typing.Any, Ellipsis)
+    # This solution should hopefully be general for *any* type that allows
+    # tuples in annotations
+    fmt = [format_annotation(a, config) for a in t]
+    if len(fmt) == 0:
+        return "()"
+    elif len(fmt) == 1:
+        return f"({fmt[0]}, )"
+    else:
+        return f"({', '.join(fmt)})"
+
+
 def format_annotation(annotation: Any, config: Config) -> str:
     typehints_formatter: Callable[..., str] | None = getattr(config, "typehints_formatter", None)
     if typehints_formatter is not None:
@@ -101,6 +117,9 @@ def format_annotation(annotation: Any, config: Config) -> str:
         return ":py:obj:`None`"
     elif annotation is Ellipsis:
         return "..."
+
+    if isinstance(annotation, tuple):
+        return format_internal_tuple(annotation, config)
 
     # Type variables are also handled specially
     try:
@@ -150,7 +169,10 @@ def format_annotation(annotation: Any, config: Config) -> str:
         formatted_args = f"\\[{', '.join(repr(arg) for arg in args)}]"
 
     if args and not formatted_args:
-        fmt = [format_annotation(arg, config) for arg in args]
+        try:
+            fmt = [format_annotation(arg, config) for arg in args]
+        except TypeError:
+            fmt = [format_annotation(args, config)]
         formatted_args = args_format.format(", ".join(fmt))
 
     return f":py:{role}:`{prefix}{full_name}`{formatted_args}"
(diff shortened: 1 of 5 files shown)


src/sphinx_autodoc_typehints/__init__.py
```diff
@@ -150,7 +169,10 @@ def format_annotation(annotation: Any, config: Config) -> str:
         formatted_args = f"\\[{', '.join(repr(arg) for arg in args)}]"
 
     if args and not formatted_args:
-        fmt = [format_annotation(arg, config) for arg in args]
+        try:
+            fmt = [format_annotation(arg, config) for arg in args]
+        except TypeError:
```
bskinn, line 174: Also not entirely sure if this exception handling needs to be refined.

src/sphinx_autodoc_typehints/__init__.py
```diff
@@ -89,6 +89,22 @@ def get_annotation_args(annotation: Any, module: str, class_name: str) -> tuple[
     return getattr(annotation, "__args__", ())
 
 
+def format_internal_tuple(t: tuple, config: Config) -> str:
+    # An annotation can be a tuple, e.g., for nptyping:
+    #   NDArray[(typing.Any, ...), Float]
+    # In this case, format_annotation receives:
+    #   (typing.Any, Ellipsis)
+    # This solution should hopefully be general for *any* type that allows
+    # tuples in annotations
```
gaborbernat, line 98: Please align it alongside 120 characters

src/sphinx_autodoc_typehints/__init__.py
gaborbernat: Why do we need this?

gaborbernat:
> Prime among them is whether introducing nptyping as a test/dev dependency is a good idea.

I see it's a pure wheel so I'm ok with this.

> Also, input on whether this implementation is sufficiently general, and whether additional/different test cases should be added/used would be good.

Can be a good starting place. We can add additional tests if users report bugs, no?

src/sphinx_autodoc_typehints/__init__.py
bskinn: To handle things with bare integers in annotations, like `Float[32]` in `nptyping` and (presumably) [`Literal[640]` in PEP 646](https://www.python.org/dev/peps/pep-0646/#use-case-1-specifying-shape-values).

The `TypeError` here is raised because `args` is not iterable in these cases. Singletons are not an issue in other subscriptable types because they are special-cased in their handling earlier in the function.

I did the error trapping this way initially b/c EAFP model. I can think of two other ways to handle offhand at this point in the code, both of which are LBYL, but more precise:

1. Switch to a `try: iter(args); except ...` handler. Would make it more specifically catch the iterable/non-iterable difference.

2. Switch to an explicit `isinstance(args, Number)` check. This would be a very focused check, and would not cover cases with non-numeric singletons. I believe neither `nptyping` nor PEP 646 uses non-numeric singletons, but I'm not 100% sure; and, who knows what's out there in the ecosystem, or will be.

I think I'm inclined to revise to (1).

src/sphinx_autodoc_typehints/__init__.py
bskinn: I'm not sure what you mean -- do you want me to remove the indentation?

bskinn:
> > Prime among them is whether introducing nptyping as a test/dev dependency is a good idea.
> 
> I see it's a pure wheel so I'm ok with this.
> 

The main thing I was thinking of is, currently my test cases depend on some internal type transformations that `nptyping` does. If `nptyping` changes these transformations, that will cause this project's test suite to fail. I realized: I can just change the test examples so that `nptyping` does not transform them. That should remove the brittleness, at least most of the way. I'll make this change.

> > Also, input on whether this implementation is sufficiently general, and whether additional/different test cases should be added/used would be good.
> 
> Can be a good starting place. We can add additional tests if users report bugs, no?

Yep. That's what I was thinking.

src/sphinx_autodoc_typehints/__init__.py
gaborbernat: Sounds good for 1 👍

src/sphinx_autodoc_typehints/__init__.py
gaborbernat: yes; this line could reach all the way to 120 character, so no need to break it so soon

gaborbernat:
> That should remove the brittleness, at least most of the way. I'll make this change.

Sounds good on this.

bskinn:
For my reference:

- [ ] De-fold comment
- [ ] Re-work iterable/non-iterable check
- [ ] Fix typing problem
- [ ] Rewrite tests to avoid `nptyping` brittleness

src/sphinx_autodoc_typehints/__init__.py
bskinn: Fixed in aa53976

bskinn:
b279611 removes as much of the brittleness from the `nptyping` dependency as is possible with the current [`assert result == expected_result` check](https://github.com/tox-dev/sphinx-autodoc-typehints/blob/7b8c357be15d7b7b4ec4ad272f554f7ab3e0e197/tests/test_sphinx_autodoc_typehints.py#L209).

If `nptyping` restructures/refactors their internals, these test cases will break.

It would be possible to reduce/eliminate this risk of breakage by loosening the assert, perhaps only for the `nptyping` cases; e.g.,:

```

if "nptyping" in result:
    # Looser check that matches e.g. re.search(expected_regex, result)
    # where expected_regex has things in it like r"~nptyping\..+\.NDArray"
else:
    assert result == expected_result
```

That would make the test function messy, though.

I guess another alternative would be to make a new test function for the nptyping cases, with the above selective-matching behavior.

-----

TLDR; are you okay with accepting this possibility for breakage from `nptyping` for now, and we'll see what happens? Or, would you prefer to try to mitigate it now?

gaborbernat (approved):

gaborbernat (changes requested):
Please add a changelog with 1.16.0

https://github.com/tox-dev/sphinx-autodoc-typehints/blob/main/CHANGELOG.md

bskinn:
> Please add a changelog with 1.16.0
> 
> https://github.com/tox-dev/sphinx-autodoc-typehints/blob/main/CHANGELOG.md

Done!

gaborbernat (approved):

PR #218, opened 2022-02-11
Add `typehints_use_rtype` option

### LINKED ISSUE #217: Playing nicely with `napoleon_use_rtype`
Hi,
Napoleon provides the [napoleon_use_rtype](https://www.sphinx-doc.org/en/master/usage/extensions/napoleon.html#confval-napoleon_use_rtype) option, which can be set to `False`, such that the return type is documented "inline" with the "Returns" section, rather than creating a separate section for it.

However, this extension ignores this (or any presence of a type documented in the `:returns:` directive) and creates an `:rtype:` directive. This may lead to duplicate rendering of the return types.

I have a small fix for this. It is adding about 5 lines of code, and a new option named `typehints_use_rtype`, which can be used in conjunction with `napoleon_use_rtype` (or hand written `:return:` directives of similar format).

If you think this is something you would like to include in this extension I am happy to open a PR. Please let me know.

Simon

SimonHeybrock:
This fixes an interoperability issue with `sphinx.ext.napoleon`. See issue and code for details.

There is a string-matching aspect to this (see comment in code) that I am not happy with. Is there a better solution? Note that this potentially brittle behavior would only occur if this new option is explicitly disabled, i.e., it should not break existing behavior.

Fixes #217.

### DIFF
--- a/src/sphinx_autodoc_typehints/__init__.py
+++ b/src/sphinx_autodoc_typehints/__init__.py
@@ -556,13 +556,22 @@ def _inject_types_to_docstring(
                 insert_index = None
                 break
             elif line.startswith(":return:") or line.startswith(":returns:"):
+                # This is brittle, I have not found information on how standard the use of " -- " to
+                # separate type from description is.
+                if " -- " in line and not app.config.typehints_use_rtype:
+                    insert_index = None
+                    break
                 insert_index = at
 
         if insert_index is not None and app.config.typehints_document_rtype:
             if insert_index == len(lines):  # ensure that :rtype: doesn't get joined with a paragraph of text
                 lines.append("")
                 insert_index += 1
-            lines.insert(insert_index, f":rtype: {formatted_annotation}")
+            if app.config.typehints_use_rtype or insert_index == len(lines):
+                lines.insert(insert_index, f":rtype: {formatted_annotation}")
+            else:
+                s = lines[insert_index]
+                lines[insert_index] = f":return: {formatted_annotation} --{s[s.find(' '):]}"
 
 
 def validate_config(app: Sphinx, env: BuildEnvironment, docnames: list[str]) -> None:  # noqa: U100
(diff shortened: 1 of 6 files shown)


src/sphinx_autodoc_typehints/__init__.py
```diff
@@ -556,13 +556,22 @@ def _inject_types_to_docstring(
...
 
         if insert_index is not None and app.config.typehints_document_rtype:
             if insert_index == len(lines):  # ensure that :rtype: doesn't get joined with a paragraph of text
                 lines.append("")
                 insert_index += 1
-            lines.insert(insert_index, f":rtype: {formatted_annotation}")
+            if app.config.typehints_use_rtype or insert_index == len(lines):
+                lines.insert(insert_index, f":rtype: {formatted_annotation}")
+            else:
+                s = lines[insert_index]
```
gaborbernat, line 573: Single char variables should be only used inside comrehensions.

src/sphinx_autodoc_typehints/__init__.py
```diff
@@ -556,13 +556,22 @@ def _inject_types_to_docstring(
                 insert_index = None
                 break
             elif line.startswith(":return:") or line.startswith(":returns:"):
+                # This is brittle, I have not found information on how standard the use of " -- " to
+                # separate type from description is.
+                if " -- " in line and not app.config.typehints_use_rtype:
```
gaborbernat, line 561: I'd like something better than this 🤔

gaborbernat (changes requested):
You'll need to add test and changelog 🤔

src/sphinx_autodoc_typehints/__init__.py
gaborbernat: Remove the comment and we're good.

SimonHeybrock: I also added a test for this case (using `Union[int, str]` as an example).

gaborbernat (approved):

SimonHeybrock:
Thanks! Though we forgot to edit the changelog in the end!

gaborbernat:
https://github.com/tox-dev/sphinx-autodoc-typehints/releases/tag/1.17.0

PR #225, opened 2022-04-08
Fix mock imports on guarded imports

khameeteman:
This PR solves an issue where a mocked module is imported in a module, which is in turn a type guarded import:

module_b:
```python
from viktor import AI  # 'viktor' part of autodoc_mock_imports
...
```

module_a:
```python
if TYPE_CHECKING:
    from module_b import ...
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Changelog
 
+## 1.18.1
+
+- Fix mocked module import not working when used as guarded import
+
 ## 1.18.0
 
 - Support and require `nptyping>=2`
--- a/src/sphinx_autodoc_typehints/__init__.py
+++ b/src/sphinx_autodoc_typehints/__init__.py
@@ -5,13 +5,14 @@ import re
 import sys
 import textwrap
 from ast import FunctionDef, Module, stmt
-from typing import _eval_type, List  # type: ignore # no import defined in stubs
-from typing import Any, AnyStr, Callable, ForwardRef, NewType, TypeVar, get_type_hints
+from typing import _eval_type  # type: ignore # no import defined in stubs
+from typing import Any, AnyStr, Callable, ForwardRef, NewType, TypeVar, get_type_hints, List
 
 from sphinx.application import Sphinx
 from sphinx.config import Config
 from sphinx.environment import BuildEnvironment
-from sphinx.ext.autodoc import Options, mock
+from sphinx.ext.autodoc import Options
+from sphinx.ext.autodoc.mock import mock
 from sphinx.util import logging
 from sphinx.util.inspect import signature as sphinx_signature
 from sphinx.util.inspect import stringify_signature
--- a/src/sphinx_autodoc_typehints/__init__.py
+++ b/src/sphinx_autodoc_typehints/__init__.py
@@ -11,7 +11,7 @@ from typing import Any, AnyStr, Callable, ForwardRef, NewType, TypeVar, get_type
 from sphinx.application import Sphinx
 from sphinx.config import Config
 from sphinx.environment import BuildEnvironment
-from sphinx.ext.autodoc import Options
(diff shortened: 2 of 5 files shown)


tests/roots/test-resolve-typing-guard/demo_typing_guard_dummy.py
```diff
@@ -0,0 +1,5 @@
+from viktor import AI  # 'viktor' part of autodoc_mock_imports
```
gaborbernat, line 1: What is autodoc_mock_imports?

gaborbernat: Ping on this :blush:

khameeteman: External dependencies can be mocked when specified in the conf.py: https://www.sphinx-doc.org/en/master/usage/extensions/autodoc.html#confval-autodoc_mock_imports

gaborbernat (changes requested):
You'll need to pass the CI.

khameeteman:
> You'll need to pass the CI.

Ye I think it is failing because nptyping has recently released a new major 2.0.0

gaborbernat:
Can you put in another PR to fix that first then? :blush:

src/sphinx_autodoc_typehints/__init__.py
```diff
@@ -334,13 +334,14 @@ def _resolve_type_guarded_imports(obj: Any) -> None:
                     for (_, part) in _TYPE_GUARD_IMPORT_RE.findall(module_code):
                         guarded_code = textwrap.dedent(part)
                         try:
-                            exec(guarded_code, obj.__globals__)
+                            with mock(app.config.autodoc_mock_imports):
```
gaborbernat, line 337: Let's pass in `app.config.autodoc_mock_imports` as argument and not the entire `app` object.

tests/roots/test-resolve-typing-guard/demo_typing_guard_dummy.py
```diff
@@ -0,0 +1,5 @@
...
+
+
+class AnotherClass:
+    pass
```
gaborbernat, line 5: use ellipsis instead of pass

gaborbernat (changes requested):

tests/roots/test-resolve-typing-guard/demo_typing_guard_dummy.py
gaborbernat: Please also rebase on main and add a changelog entry 👍

CHANGELOG.md
```diff
@@ -1,5 +1,7 @@
 # Changelog
 
+- Fix mocked module import not working when used as guarded import
```
gaborbernat, line 3: You'll need to fix the merge conflic here and use version 1.18.1

khameeteman: Updated

src/sphinx_autodoc_typehints/__init__.py
```diff
@@ -303,24 +303,24 @@ def _future_annotations_imported(obj: Any) -> bool:
     return bool(_annotations.compiler_flag == future_annotations)
 
 
-def get_all_type_hints(obj: Any, name: str) -> dict[str, Any]:
-    result = _get_type_hint(name, obj)
+def get_all_type_hints(autodoc_mock_imports: List[str], obj: Any, name: str) -> dict[str, Any]:
```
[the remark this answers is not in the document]
khameeteman: Not sure, here it said `List[str]`: https://github.com/sphinx-doc/sphinx/blob/5.x/sphinx/ext/autodoc/mock.py#L133

gaborbernat (approved):

PR #248, opened 2022-07-29
Add support for recursive type hints

### LINKED ISSUE #223: Recursion errors with some recursive types
It appears #91 has been reintroduced by the 1.6.0 release. That is, a recursive type definition like

```py
TreeType = typing.Dict[str, typing.Union['TreeType', typing.List['TreeType']]]
```

causes documentation generation to fail with the following recursion error.

```py
Extension error (sphinx_autodoc_typehints):
Handler <function process_docstring at 0x7ff706291310> for event 'autodoc-process-docstring' threw an exception (exception: maximum recursion depth exceeded)
```

hmc-cs-mdrissi:
Closes #223. This is simple approach that uses string of forward reference and avoids needing to evaluate recursive/mutually recursive type aliases.

I've added two test cases that use recursive aliases and previously would crash. I've also been using this fork for a while in company internal codebase and it's worked well for our recursive annotations.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Changelog
 
+## 1.19.1
+
+- Fix bug for recursive type alias.
+
 ## 1.19.0
 
 - Support for CPython 3.11, no longer adds `Optional` when the argument is default per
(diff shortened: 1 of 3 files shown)


gaborbernat (commented):
Can you fix the type checker and add a changelog entry? Thanks!

hmc-cs-mdrissi:
Fixed. Thank you for the fast review.

gaborbernat (approved):

Issue #237, opened 2022-06-14
data role used for collections.abc.Callable while class is required


Numerlor:
The current code forces the data role for `callable` regardless of the module it's defined in, causing incorrect symbol references to be created if the module is `collections.abc.Callable`.

Looking at the code it may also be missing some special handling in `format_annotation` as there's a check for the full `typing.Callable` name that skips the collections generic

Numerlor:
@gaborbernat I'm willing to pick this up with possibly some of the other issues as the code seems relatively simple, but I want to make sure someone would be around to review/merge if I do

gaborbernat:
![hello](https://media.tenor.com/images/43b8ed43e6ac683091db7c8314757923/tenor.gif)

PR #250, opened 2022-08-06
only convert role for pydata annotations if module is typing

Numerlor:
The names may have an origin that's not typing, causing an incorrect
role to be applied to them

fixes: #237


gaborbernat (changes requested):
Can you please add a changelog entry?

gaborbernat (approved):

Issue #253, opened 2022-09-12
Support variadic arguments (*args) and keyword arguments (**kwargs)


Bruntaz:
Using Sphinx and Napoleon on this code: 

```python
class MyClass:
    def my_func(self, a: int, b: str, *c: bool, **d: float):
        """
        Args:
            a: Some description for a
            b: Some description for b
            *c: Some description for c
            **d: Some description for d
        """
```

produces the output:
![image](https://user-images.githubusercontent.com/15197528/189658977-ad5cc523-79a1-4545-ba43-9306e67d8d7a.png)

sphinx-autodoc-typehints doesn't like the `*c` or `**d` syntax in the docstring though, and treats it as if there is no type for those parameters:
![image](https://user-images.githubusercontent.com/15197528/189658770-859cec67-8584-4672-ad0a-0b9e1f4fa711.png)

This can be worked around by not using `*`s in the docstring, and just using `c` and `d` instead:

```python
class MyClass:
    def my_func(self, a: int, b: str, *c: bool, **d: float):
        """
        Args:
            a: Some description for a
            b: Some description for b
            c: Some description for c
            d: Some description for d
        """
```
![image](https://user-images.githubusercontent.com/15197528/189659128-3c6a1572-93f4-40f3-8245-7dca3e897788.png)

I think that sphinx-autodoc-typehints should allow the `*` syntax, since Sphinx/Napoleon accepts it as correct and it makes it explicit in the argument description that the arg is actually N arguments or keyword arguments. 

I would (naively?) assume this is a relatively small fix, to either lstrip `*`s from args before looking them up by name.

gaborbernat:
A PR for this would be welcome.

PR #254, opened 2022-09-18
allow starred args and kwargs in docstring params

Numerlor:
closes: #253


Numerlor:
Should the changelog go under 1.20?

I'm also not quite sure on the 3.7 test, as the part I added is without differences in the assert

gaborbernat (commented):
Yeah 1.20 is fine.

PR #255, opened 2022-09-18
Add support for paramspec

Numerlor:
Adds support for ParamSpec, to keep compat I've decided to only check the name as typing_extensions is not a normal dep, and someone else using the name seems unlikely, though a check for some attribute could be added if necessary

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@
 ## 1.20
 
 - Use hatchling instead of setuptools
+- Add support for typing.ParamSpec
 
 ## 1.19.2
 
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -28,7 +28,7 @@ optional-dependencies.testing = [
   "sphobjinv>=2.2.2",
   "typing-extensions>=4.3",
 ]
-optional-dependencies.type_comment = ['typed-ast>=1.5.4; python_version < "3.8"']
+optional-dependencies.type-comment = ['typed-ast>=1.5.4; python_version < "3.8"']
 dynamic = ["version"]
 classifiers = [
   "Development Status :: 5 - Production/Stable",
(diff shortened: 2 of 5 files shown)


gaborbernat (changes requested):
Please add test, changelog  entry.

gaborbernat (changes requested):
Some of the checks still fail.

Numerlor:
Not sure about the 3.7 check, as it's failing for me on main too

gaborbernat:
> Not sure about the 3.7 check, as it's failing for me on main too, and type is something from stubs?

Indeed something changed between https://github.com/tox-dev/sphinx-autodoc-typehints/actions/runs/3043415433 and https://github.com/tox-dev/sphinx-autodoc-typehints/actions/runs/3051184692. 

The type check failure is because of because of [github.com/python/mypy/pull/13500](https://github.com/python/mypy/pull/13500) (see related [github.com/python/mypy/issues/13627](https://github.com/python/mypy/issues/13627)); waiting for mypy 0.980 release to fix that. I don't have currently the bandwith to look into this, so if you want this to land soon would be much apprecated if you can look into it and put in a fix PR.

Numerlor:
Okay, for 3.7, it looks like the type_comment extra is not getting installed and type comments are skipped because of that.
The trailing ``s used here https://github.com/tox-dev/sphinx-autodoc-typehints/blob/ccc75d244d0f4452048f80fc73e39a34a4eefc00/tox.ini#L25 was dropped in ccc75d244d0f4452048f80fc73e39a34a4eefc00, but after removing it in tox.ini, it still is unable to find it unless the underscore is removed or changed to `-`, not sure if that's a bug somewhere or the intended behaviour.

Should I just change it to `-`?

gaborbernat:
> Should I just change it to `-`?

Can give it a try to see if helps.

PR #278, opened 2023-01-04
Added options to retain original typehints in signatures

BhavyeMathur:
I was using this library and noticed that there was no existing way to retain the typehints automatically generated by autodoc in function/class/method signatures. I quite liked the way that return types are annotated with the -> in signatures and wanted to preserve those while using autodoc-typehints.

This pull request adds 2 simple options that **do not alter the behaviour of anyone already using typehints**:

1. `typehints_use_signature` (`False` by default)
2. `typehints_use_signature_return` (`False` by default)

By setting these, the `sph_signature.replace()` commands in `process_signature()` can be altered

With default options:
image

With the new options set to `True`:
image

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -6,4 +6,3 @@ dist
 .*_cache
 /src/sphinx_autodoc_typehints/version.py
 venv*
-.idea*
--- a/.gitignore
+++ b/.gitignore
@@ -6,3 +6,4 @@ dist
 .*_cache
 /src/sphinx_autodoc_typehints/version.py
 venv*
+.idea*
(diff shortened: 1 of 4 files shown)


gaborbernat (changes requested):
Tests, please!

BhavyeMathur:
Hey! So I'm new to pull requests and to sphinx. Any particular tests you're looking for?

gaborbernat:
Test for all features you've added.

.gitignore
```diff
@@ -6,3 +6,4 @@ dist
 .*_cache
 /src/sphinx_autodoc_typehints/version.py
 venv*
+.idea*
```
gaborbernat, line 9: Remove this, IDE ignore should be in your global gitignore not per project.

gaborbernat (changes requested):

gaborbernat (approved):

PR #285, opened 2023-01-13
Set role to data when creating an Optional from `| None`

hoodmane:
This is not super important, but when converting `| None` to `Optional` the code sets the role incorrectly to `class` and so generates dead links.


gaborbernat (changes requested):
Please add tests 👍 I think is data only on 3.11 not 3.10.

hoodmane:
It's also a data in Python 3.10 -- note that `Optional` is in `_PYDATA_ANNOTATIONS` so if an `Optional` is generated directly instead of via a conversion from `A | None` then it will get `:py:data:`. So the test can just ask for e.g., `Union[str, None]` and `Optional[str]` to produce the same output.

PR #287, opened 2023-01-14
Remove redundant return type for attributes

### LINKED ISSUE #286: Property :rtype: is redundant
Looking at the output, it seems that properties have redundant types. In the test suite, `a_property` is an example. Defined like:
```py
@property
def a_property(self) -> str:
    """
    Property docstring
    """
```
The test suite asserts that the type of the property is included in two places. This strikes me as redundant and slightly confusing: does a property really "return" a value? In my opinion, the abstraction makes it look like we are "accessing" the value rather than returning it. To be specific, I want the following change in the output.
```patch
     property a_property: str
  
        Property docstring
  
-       Return type:
-          "str"
- 
```
I applied the following patch to do this:
```patch
--- a/src/sphinx_autodoc_typehints/__init__.py
+++ b/src/sphinx_autodoc_typehints/__init__.py
@@ -611,7 +611,7 @@ def _inject_types_to_docstring(
 
                 lines.insert(insert_index, type_annotation)
 
-    if "return" in type_hints and not inspect.isclass(original_obj):
+    if "return" in type_hints and not inspect.isclass(original_obj) and not inspect.isdatadescriptor(original_obj):
         if what == "method" and name.endswith(".__init__"):  # avoid adding a return type for data class __init__
             return
         formatted_annotation = format_annotation(type_hints["return"], app.config)

 ```
Do people prefer the current behavior? If some people want the redundant property types, could we add a config option to opt out of it?

### PR DESCRIPTION
**Pull request by hoodmane**
Resolves #286.

### DIFF
--- a/src/sphinx_autodoc_typehints/__init__.py
+++ b/src/sphinx_autodoc_typehints/__init__.py
@@ -611,11 +611,7 @@ def _inject_types_to_docstring(
 
                 lines.insert(insert_index, type_annotation)
 
-    if (
-        "return" in type_hints
-        and not inspect.isclass(original_obj)
-        and not inspect.isdatadescriptor(original_obj)  # noqa: SC200 # Possibly misspelt word: 'isdatadescriptor'
-    ):
+    if "return" in type_hints and not inspect.isclass(original_obj) and not inspect.isdatadescriptor(original_obj):
         if what == "method" and name.endswith(".__init__"):  # avoid adding a return type for data class __init__
             return
         formatted_annotation = format_annotation(type_hints["return"], app.config)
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

File: src/sphinx_autodoc_typehints/__init__.py [hunk 1]
```diff
@@ -609,7 +609,11 @@ def _inject_types_to_docstring(
 
                 lines.insert(insert_index, type_annotation)
 
-    if "return" in type_hints and not inspect.isclass(original_obj):
+    if (
+        "return" in type_hints
+        and not inspect.isclass(original_obj)
+        and not inspect.isdatadescriptor(original_obj)  # noqa: SC200 # Possibly misspelt word: 'isdatadescriptor'
```
gaborbernat, line 615: Add this word to whitelist.txt instead.

gaborbernat (changes requested):

src/sphinx_autodoc_typehints/__init__.py
hoodmane: Okay. We also need to add `getitem` there...

gaborbernat (approved):

PR #294, opened 2023-01-16
Fix spacing between `:rtype:` and directives

### LINKED ISSUE #293: Invalid spacing between `:rtype:` and following directives
In the 1.21.0 release, the position of the emitted `:rtype:` field seems to have changed so that it is now earlier in the docstring, rather than reliably being at the end.  This is fine, except when the place the injector chooses to place it is immediately before a `.. <whatever>::` directive. There doesn't seem to reliable be a blank line inserted after it, and if not, Sphinx complains that the field list ends unexpectedly.

For example, given this Python file `tmp/__init__.py`:
```python
def my_function() -> None:
    """
    .. rubric:: Notes

    Hello, world.
    """
```
and this Sphinx `conf.py`:
```python
project = 'tmp'
extensions = [
    "sphinx.ext.autodoc",
    "sphinx_autodoc_typehints",
]
```
running Sphinx's `make html` gives:
```text
Running Sphinx v6.1.3
loading pickled environment... done
building [mo]: targets for 0 po files that are out of date
writing output...
building [html]: targets for 0 source files that are out of date
updating environment: 0 added, 1 changed, 0 removed
reading sources... [100%] index
/Users/jake/tmp/autodoc-typehints/pkg/tmp/__init__.py:docstring of tmp.my_function:1: WARNING: Field list ends without a blank line; unexpected unindent.
looking for now-outdated files... none found
pickling environment... done
checking consistency... done
preparing documents... done
writing output... [100%] index
generating indices... genindex py-modindex done
writing additional pages... search done
copying static files... done
copying extra files... done
dumping search index in English (code: en)... done
dumping object inventory... done
build succeeded, 1 warning.
```

The docstring enters `_inject_types_to_docstring` as
```text
.. rubric:: Notes

Hello, world.

```
and is modified to
```text
:rtype: :py:obj:`None`
.. rubric:: Notes

Hello, world.

```

This also happens if there's any regular text before the `.. rubric:: notes` as well (which is a bit more of a realistic scenario).

jakelishman:
If the `:rtype:` field is being inserted immediately before a directive, we must ensure there is a blank line between the two to avoid a Sphinx warning.

Given that #290, which added this behaviour, already added a suitable test that emits a warning that doesn't fail the suite, I wasn't clear on how best to add a further test for this.

Fix #293.

### DIFF
--- a/src/sphinx_autodoc_typehints/__init__.py
+++ b/src/sphinx_autodoc_typehints/__init__.py
@@ -633,6 +633,7 @@ def _inject_types_to_docstring(
             return
         formatted_annotation = format_annotation(type_hints["return"], app.config)
         insert_index = len(lines)
+        extra_newline = False
         for at, line in enumerate(lines):
             if line.startswith(":rtype:"):
                 insert_index = None
@@ -643,8 +644,9 @@ def _inject_types_to_docstring(
                     break
                 insert_index = at
             elif line.startswith(".."):
-                # Make sure that rtype comes before any usage or examples section
+                # Make sure that rtype comes before any usage or examples section, with a blank line between.
                 insert_index = at
+                extra_newline = True
                 break
 
         if insert_index is not None and app.config.typehints_document_rtype:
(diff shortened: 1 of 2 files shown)


jakelishman:
I'm not clear why the docstring complexity flake8 is failing here, when I haven't touched the docstring.

gaborbernat (changes requested):
Test and make the CI pass.

gaborbernat:
cc @hoodmane

jakelishman:
I added an assertion about the warning within the current testing setup, though it feels like a fairly fragile test. I'd generally have preferred to just have all warnings treated as errors (except explicit overrides) within that test, but there are other warnings that would be triggered, and I'm not so familiar with your testing structure to make that change.

To satisfy the complexity checker I had to completely split the containing function, rather than the more minimal change that's only in the first commit.  The logical changeset is the same as 3966fb3, but it's hard to see from the diff.

hoodmane:
+1 for whitelisting warnings to allow rather than blacklisting warnings to not allow.

gaborbernat (approved):

PR #299, opened 2023-01-17
Use format_annotation to render class attribute type annotations

### LINKED ISSUE #298: Does not handle class attributes
I have a fix for this but it isn't great. More monkey patching. I'll open a PR and provide an explanation there.

Suppose I have the following:

```py
class T:
    """A class"""

    code : CodeType | None
    """An attribute"""
```
#### What I see:

The type for the attribute was rendered directly by `sphinx-autodoc` inconsistently with the way that `sphinx-autodoc-typehints` would render it:

Screenshot from 2023-01-17 10-16-18

#### What I'd like to see:

The attribute type rendered by `sphinx-autodoc-typehints` 

Screenshot from 2023-01-17 10-16-59

hoodmane:
Resolves #298.

This is a bit hard to do because autodoc doesn't emit any events to help us adjust the type rendering for attributes and because the type annotation is currently converted to text and then rendered from text by the `PyAttribute` directive.

The system in autodoc for attribute rendering is:
1. In `AttributeDocumenter.get_directive_header`, call `stringify_annotation` to render the annotation to a string
2. add `:type: stringified annotation` to the directive header of `PyAttribute`
3. When `sphinx.domains.python.PyAttribute` sees the `:type:` it calls `format_annotation` which attempts to convert the string representation of the type into a more fancy representation with cross referencing

Notes:
2. we don't get an opportunity to adjust the directive header with `autodoc-process-docstring` or some other event
3. there isn't enough information retained in the string for `_inject_types_to_docstring` to work.

So my approach is:
1. Patch `AttributeDocumenter.get_directive_header` to use `format_annotation` instead of `stringify_annotation`. Add a prefix to the output to indicate that it is rst not a plain text type annotation.
3. Patch `PyAttribute` so that when it sees the prefix, it formats the output as rst rather than trying to handle it as a plain text type annotation.

As future proofing, we check that the patch targets exist and bail if they don't exist. If they move this patch should gracefully decay into a no-op rather than crashing.

### DIFF
--- a/src/sphinx_autodoc_typehints/attributes_patch.py
+++ b/src/sphinx_autodoc_typehints/attributes_patch.py
@@ -18,7 +18,7 @@ _parse_annotation = getattr(sphinx.domains.python, "_parse_annotation", None)
 # We want to patch:
 # * sphinx.ext.autodoc.stringify_typehint (in sphinx < 6.1)
 # * sphinx.ext.autodoc.stringify_annotation (in sphinx >= 6.1)
-STRINGIFY_PATCH_TARGET = None
+STRINGIFY_PATCH_TARGET = ""
 for target in ["stringify_typehint", "stringify_annotation"]:
     if hasattr(sphinx.ext.autodoc, target):
         STRINGIFY_PATCH_TARGET = f"sphinx.ext.autodoc.{target}"
--- a/src/sphinx_autodoc_typehints/attributes_patch.py
+++ b/src/sphinx_autodoc_typehints/attributes_patch.py
@@ -76,8 +76,10 @@ def patched_parse_annotation(settings: Values, typ: str, env: Any) -> Any:
 
 def patch_py_attribute_handle_signature() -> None:
     """
-    Patch PyAttribute.handle_signature so that it treats the :type: as rst and
-    renders it as-is rather than trying to parse it as a type annotation string.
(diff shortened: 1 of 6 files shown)


gaborbernat (approved):

src/sphinx_autodoc_typehints/attributes_patch.py
```diff
@@ -0,0 +1,102 @@
...
+from sphinx.domains.python import PyAttribute
+from sphinx.ext.autodoc import AttributeDocumenter
+
+# Defensively check for the things we want to patch
+_parse_annotation = getattr(sphinx.domains.python, "_parse_annotation", None)
+
+# We want to patch:
+# * sphinx.ext.autodoc.stringify_typehint (in sphinx < 6.1)
+# * sphinx.ext.autodoc.stringify_annotation (in sphinx >= 6.1)
+STRINGIFY_PATCH_TARGET = ""
```
gaborbernat, line 21: You could use none just need to annotation it `str | None`, no?

hoodmane: That makes a type error when I try to use it unless I add an extra assert. Empty string is falsey and this makes the type checker happy without adding the assert.

gaborbernat (approved):

PR #307, opened 2023-01-20
Resolve issue 305

### LINKED ISSUE #305: WARNING: Field list ends without a blank line; unexpected unindent
The `WARNING: Field list ends without a blank line; unexpected unindent` has returned in 1.21.5. I suspect this is another variation on the problem reported and fixed in #293.

GitHub Actions log showing this failure: https://github.com/lsst-sqre/safir/actions/runs/3963428111/jobs/6791355176

Example docstring that produces the first error message there:

```python
    """Run the decorated function with `asyncio.run`.

    Intended to be used as a decorator around an async function that needs to
    be run in a sync context.  The decorated function will be run with
    `asyncio.run` when invoked.  The caller must not already be inside an
    asyncio task.

    Parameters
    ----------
    f
        The function to wrap.

    Examples
    --------
    An application that uses Safir and `Click`_ may use the following Click
    command function to initialize a database.

    .. code-block:: python

       import structlog
       from safir.asyncio import run_with_asyncio
       from safir.database import initialize_database

       from .config import config
       from .schema import Base


       @main.command()
       @run_with_asyncio
       async def init() -> None:
           logger = structlog.get_logger(config.safir.logger_name)
           engine = await initialize_database(
               config.database_url,
               config.database_password,
               logger,
               schema=Base.metadata,
           )
           await engine.dispose()
    """
```

Removing the `Parameters` section makes the warning go away. The same docstring (and the others in this package) worked correctly with 1.21.1.

hoodmane:
Apparently we need to subtract 3 here?

Resolves #305.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Changelog
 
+## [Unreleased]
+
+- Fix a `Field list ends without a blank line` warning (issue 305).
+
 ## 1.21.5
 
 - More robust determination of rtype location / fix issue 302
--- a/src/sphinx_autodoc_typehints/__init__.py
+++ b/src/sphinx_autodoc_typehints/__init__.py
@@ -716,7 +716,7 @@ def get_insert_index(app: Sphinx, lines: list[str]) -> InsertIndexInfo | None:
         # at end. (I don't know of any input where this happens.)
         next_sibling = child.next_node(descend=False, siblings=True)
         line_no = node_line_no(next_sibling) if next_sibling else None
-        at = line_no - 3 if line_no else len(lines)
+        at = line_no - 2 if line_no else len(lines)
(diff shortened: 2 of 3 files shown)


gaborbernat (changes requested):
Can you add a changelog entry please? https://github.com/tox-dev/sphinx-autodoc-typehints/blob/main/CHANGELOG.md#1215

hoodmane:
Done.

CHANGELOG.md
```diff
@@ -1,5 +1,9 @@
 # Changelog
 
+## [Unreleased]
```
gaborbernat, line 3: Would have helped to put here 1.21.6; now I need to do that change myself. In future please do so.

hoodmane: Okay!

PR #310, opened 2023-01-20
NFC Add a flake8 exclude rule to .pre-commit

hoodmane:


.pre-commit-config.yaml
```diff
@@ -47,6 +47,7 @@ repos:
     rev: 6.0.0
     hooks:
       - id: flake8
+        exclude: ^tests/test_integration.py
```
gaborbernat, line 50: I don't want to exclude this 🤔 Why would we do that?

hoodmane: There's just a lot of intentionally unused arguments. Maybe we can disable U100 for it. Or not I can just close this.

PR #311, opened 2023-01-20
Fix napoleon handling of numpy docstrings with no explicitly provided return type

hoodmane:
If there is no explicitly provided return type, napoleon will incorrectly parse the entire return type explanation. It can be fixed by inserting a line with a single colon at the beginning of the returns section.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,12 @@
 # Changelog
 
+## 1.21.7
+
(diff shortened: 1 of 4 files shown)


gaborbernat (changes requested):
Changelog please.

hoodmane:
Done.

gaborbernat (approved):

PR #317, opened 2023-01-26
Enable parallel write

marscher:
in #9 parallel reading was enabled. Now lets do the same for writing. I've tested this with my (large) project locally. The output looks sane and the total build time was cut in half.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,6 @@
 # Changelog
 
-## 1.22 
+## 1.22
 
 - Allow Sphinx explicitly to write in parallel.
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,6 @@
 # Changelog
 
-## unreleased
+## 1.22 
 
 - Allow Sphinx explicitly to write in parallel.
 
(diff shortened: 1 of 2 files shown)


gaborbernat (changes requested):
Can we add a changelog entry for this?

marscher:
From my observation it does not, otherwise I could not explain why the build time decreases to 50%. This was taken with Sphinx 5.3

hoodmane:
Well the change is either harmless or useful so it isn't a big deal to merge it. But the code for v5.3.0 also seems to show that this defaults True:
```
         self.parallel_write_safe = kwargs.pop('parallel_write_safe', True)
```
https://github.com/sphinx-doc/sphinx/blob/v5.3.0/sphinx/extension.py#L33

used here:
https://github.com/sphinx-doc/sphinx/blob/v5.3.0/sphinx/application.py?plain=1#L1278

CHANGELOG.md
```diff
@@ -1,5 +1,9 @@
 # Changelog
 
+## unreleased
```
gaborbernat, line 3: 1.22

marscher: there you go.

gaborbernat (approved):

PR #319, opened 2023-01-26
FIX TypeError: unhashable type

### LINKED ISSUE #318: Event 'autodoc-process-docstring' threw an exception (exception: unhashable type: 'ParamSpecArgs')
Autodoc 1.21.8 fails for a ParamSpec

```python
HINT_CB_P = ParamSpec('HINT_CB_P')
HINT_CB: TypeAlias = Callable[HINT_CB_P, Any]


class MyClass):
    def func(self, callback: HINT_CB, *args: HINT_CB_P.args, **kwargs: HINT_CB_P.kwargs):
        pass
```

Error:
```
    if annotation in _TYPES_DICT:
TypeError: unhashable type: 'ParamSpecArgs'
```

https://github.com/tox-dev/sphinx-autodoc-typehints/blob/d9c4a06b654f4415d61e5ce23f46c8356a6ab92a/src/sphinx_autodoc_typehints/__init__.py#L39

spacemanspiff2007:
Attempt to fix #318.
The access to `_TYPES_DICT` should be wrapped so the error for unhashable type is captured.
This is just a quick hack that fixes the issue for me and there sure is a more elegant way of doing it ...

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,8 @@
 
 - Fixed napoleon handling of numpy docstrings with no specified return type.
 
+- Fixed crash when documenting ParamSpecArgs
+
 ## 1.21.6
 
 - Fix a `Field list ends without a blank line` warning (issue 305).
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@
 ## 1.22
 
 - Allow Sphinx explicitly to write in parallel.
+- Fixed crash when documenting ParamSpecArgs
 
 ## 1.21.7
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,8 +12,6 @@
 
 - Fixed napoleon handling of numpy docstrings with no specified return type.
 
-- Fixed crash when documenting ParamSpecArgs
-
 ## 1.21.6
 
 - Fix a `Field list ends without a blank line` warning (issue 305).
--- a/src/sphinx_autodoc_typehints/__init__.py
+++ b/src/sphinx_autodoc_typehints/__init__.py
@@ -7,7 +7,7 @@ import textwrap
 import types
 from ast import FunctionDef, Module, stmt
 from dataclasses import dataclass
-from typing import Any, AnyStr, Callable, ForwardRef, NewType, Optional, TypeVar, get_type_hints
+from typing import Any, AnyStr, Callable, ForwardRef, NewType, TypeVar, get_type_hints
 
 from docutils.frontend import OptionParser
 from docutils.nodes import Node
--- a/src/sphinx_autodoc_typehints/__init__.py
+++ b/src/sphinx_autodoc_typehints/__init__.py
@@ -49,7 +49,11 @@ def get_annotation_module(annotation: Any) -> str:
     if annotation is None:
         return "builtins"
     is_new_type = sys.version_info >= (3, 10) and isinstance(annotation, NewType)
-    if is_new_type or isinstance(annotation, TypeVar) or type(annotation).__name__ in ("ParamSpec", 'ParamSpecArgs', 'ParamSpecKwargs'):
+    if (
+        is_new_type
+        or isinstance(annotation, TypeVar)
+        or type(annotation).__name__ in ("ParamSpec", "ParamSpecArgs", "ParamSpecKwargs")
+    ):
         return "typing"
     if hasattr(annotation, "__module__"):
         return annotation.__module__  # type: ignore # deduced Any
--- a/src/sphinx_autodoc_typehints/__init__.py
+++ b/src/sphinx_autodoc_typehints/__init__.py
@@ -35,13 +35,21 @@ _TYPES_DICT = {getattr(types, name): name for name in types.__all__}
 _TYPES_DICT[types.FunctionType] = "FunctionType"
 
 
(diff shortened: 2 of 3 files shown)


spacemanspiff2007:
pre-commit automatically changed
grafik
which is imho incompatible with python < 3.10.
Is this the intended behavior?

gaborbernat (changes requested):
It needs to have a unit test and run pre-commit for stylistic fixes. Also add changelog entry.

spacemanspiff2007:
> It needs to have a unit test

I am just copying things together so some help would be appreciated.

> run pre-commit for stylistic fixes

See my question above.

src/sphinx_autodoc_typehints/__init__.py
```diff
@@ -35,13 +35,25 @@
 _TYPES_DICT[types.FunctionType] = "FunctionType"
 
 
+def _get_types_type(obj: Any) -> str | None:
+    try:
+        return _TYPES_DICT.get(obj)
+    except TypeError:
+        # e.g. exception: unhashable type
+        return None
```
hoodmane, line 43: I think the following would be a bit more explicit:
```suggestion
from collections.abc import Hashable

def _get_types_type(obj: Any) -> str | None:
    if not isinstance(obj, Hashable):
        return None
    return _TYPES_DICT.get(obj)
```

src/sphinx_autodoc_typehints/__init__.py
```diff
@@ -7,7 +7,7 @@
 import types
 from ast import FunctionDef, Module, stmt
 from dataclasses import dataclass
-from typing import Any, AnyStr, Callable, ForwardRef, NewType, TypeVar, get_type_hints
+from typing import Any, AnyStr, Callable, ForwardRef, NewType, Optional, TypeVar, get_type_hints
```
hoodmane, line 10:
```suggestion
from typing import Any, AnyStr, Callable, ForwardRef, NewType, TypeVar, get_type_hints
```

hoodmane:
> > It needs to have a unit test
>
> I am just copying things together so some help would be appreciated.

Tests look good to me.

src/sphinx_autodoc_typehints/__init__.py
spacemanspiff2007: tbh. I even thought about
```python
def _get_types_type(obj: Any) -> str | None:
    try:
        return _TYPES_DICT.get(obj)
    except:
        # e.g. exception: unhashable type
        return None
```
because does it really matter why the lookup fails? 
If the value can't be looked up it sure is not in the dictionary.

While `Hashable` might work for ParamSpec it's [not guaranteed](https://docs.python.org/3/library/collections.abc.html#id17) to work every time.
I think catching the (broad) exception would be the best solution programmatically but if you want me to use `Hashable` I'll change it accordingly

hoodmane: I usually prefer starting out by being more specific since that way we find out if our assumptions are wrong. But certainly if someone does something like:
```py
class BadClass:
    def __hash__(self):
        raise TypeError("Oops!")
```
then the class would count as hashable but attempts to hash it would always raise an error. Probably any of these things are fine.

hoodmane (approved):

tests/test_sphinx_autodoc_typehints.py
```diff
@@ -58,6 +58,8 @@
 S = TypeVar("S", bound="miss")  # type: ignore # miss not defined on purpose # noqa: F821
 W = NewType("W", str)
 P = typing_extensions.ParamSpec("P")
+P_args = P.args
+P_kwargs = P.kwargs
```
spacemanspiff2007, line 62: MyPy fails with 
```
tests/test_sphinx_autodoc_typehints.py:61: error: "object" has no attribute "args"  [attr-defined]
tests/test_sphinx_autodoc_typehints.py:62: error: "object" has no attribute "kwargs"  [attr-defined]
```

hoodmane: Add `  # type:ignore[attr-defined]` after each of these (two spaces before # one after it).

CHANGELOG.md
```diff
@@ -7,6 +7,8 @@
 
 - Fixed napoleon handling of numpy docstrings with no specified return type.
 
+- Fixed crash when documenting ParamSpecArgs
```
hoodmane, line 10: This is not the correct place for the changelog entry: 1.21.7 and 1.21.8 were already released. Though the changelog on master looks a bit weird -- there is no entry for 1.21.8 and there is an entry for 1.22.0... Maybe we can just leave it to @gaborbernat to fix it.

spacemanspiff2007: Sorry - I thought the topmost entry was the upcoming release. I didn't check the version numers 😕

hoodmane: Well you don't seem to be the only one who got confused.

hoodmane: Maybe move this into a new entry for 1.21.9 and then I'll merge it.

hoodmane:
Thanks @spacemanspiff2007!

spacemanspiff2007:
Do you have any idea when the next version will be released?

hoodmane:
> Do you have any idea when the next version will be released?

Probably in the next day or two.

PR #328, opened 2023-03-01
Fall back to the older setuptools-scm template style

### LINKED ISSUE #327: Sphinx extension loading failed (cannot import name '__version__')
I'm running into the following problem when trying to use newer versions of sphinx-autodoc-typehints.

I've narrowed it down as far as:

```
works fine = sphinx-autodoc-typehints 1.18.1
errors out = sphinx-autodoc-typehints 1.22
```

The error I'm seeing is:

```
   > Sphinx docs generation failed:
     Running Sphinx v5.3.0
     STDERR:

     Extension error:
     Could not import extension sphinx_autodoc_typehints (exception: cannot import name '__version__' from 'sphinx_autodoc_typehints.version' (/<snip>/lib/python3.10/site-packages/sphinx_autodoc_typehints/version.py))
```

The contents of `version.py` are:

```
$ cat version.py
# coding: utf-8
# file generated by setuptools_scm
# don't change, don't track in version control
version = '1.22'
version_tuple = (1, 22)
$
```

It looks like a similar report (with additional details related to isolated builds) can also be found at:

https://bugs.freebsd.org/bugzilla/show_bug.cgi?id=269686

bringhurst:
setuptools-scm changed their template style for versions.

Lets fall back to the older template style if the new style does not exist.

Resolves #327

### DIFF
--- a/src/sphinx_autodoc_typehints/__init__.py
+++ b/src/sphinx_autodoc_typehints/__init__.py
@@ -23,7 +23,13 @@ from sphinx.util.inspect import signature as sphinx_signature
 from sphinx.util.inspect import stringify_signature
 
 from .patches import install_patches
-from .version import __version__
+
+try:
+    from .version import __version__
+except ImportError:
+    # The format of this file was changed, so lets fall back to the older style.
+    # See https://github.com/tox-dev/sphinx-autodoc-typehints/issues/327
+    from .version import version as __version__
 
 _LOGGER = logging.getLogger(__name__)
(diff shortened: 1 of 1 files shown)


gaborbernat (changes requested):
Nope, please use hatch and hatch-vcs.

gaborbernat:
Specifically the versions that contain the fix and is specified in the toml file https://github.com/tox-dev/sphinx-autodoc-typehints/blob/main/pyproject.toml#L3

gaborbernat:
Good luck. Requiring modern python packaging standards is the minimum for any reasonable python project.

PR #393, opened 2023-10-24
fix: Recursively evaluate guarded code

### LINKED ISSUE #375: Resolving multiple imports guarded by typing.TYPE_CHECKING
## Hello!

At the beginning: I saw #22, but this case looks like a separate issue.

### Description

Version 1.15.0 introduced a change in the way imports guarded by `if typing.TYPE_CHECKING` are handled. The `_resolve_type_guarded_imports` function searches for `TYPE_CHECKING` occurrences using a regular expression, and then runs the code inside that block with `exec`.

This solution is pretty smart and it works, until we have multiple levels of `TYPE_CHECKING` guards.

### Example

`bokeh/model/model.py`: https://github.com/bokeh/bokeh/blob/branch-3.3/src/bokeh/model/model.py#L42

```python
if TYPE_CHECKING:
    from ..core.has_props import Setter
```

`bokeh/core/has_props.py`: https://github.com/bokeh/bokeh/blob/branch-3.3/src/bokeh/core/has_props.py#L98

```python
if TYPE_CHECKING:
    from ..client.session import ClientSession
    from ..server.session import ServerSession

if TYPE_CHECKING:
    Setter = Union[ClientSession, ServerSession]
```

The resolver finds `TYPE_CHECKING` guard in `model.py` and tries to execute the import, but it's unable to find the `Setter` in `has_props.py` (the `TYPE_CHECKING` flag is set to `False`, and the file `has_props.py` is just being executed by python, not parsed by `_resolve_type_guarded_imports` function).

### Proposition

I'm wondering if it would be possible to resolve this type of imports recursively, using `_resolve_type_guarded_imports`? Or do you have any other ideas? I guess I have to stay with version <1.15.0 until this gets fixed.

Mr-Pepe:
Partly resolves #375 

- Add test for nested guarded imports
- Handle nested type guards

### DIFF
--- a/src/sphinx_autodoc_typehints/__init__.py
+++ b/src/sphinx_autodoc_typehints/__init__.py
@@ -2,6 +2,7 @@
 from __future__ import annotations
 
 import ast
+import importlib
 import inspect
 import re
 import sys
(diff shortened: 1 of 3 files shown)


gaborbernat:
Seems it stalled.

Mr-Pepe:
It doesn't solve the Pydantic problem described in #375 but that's a Pydantic issue and I don't know if it solves @lchojnacki's problem described in #375 because they didn't respond. 

But the PR is good to go from my perspective. The code does what it's supposed to do (handle nested `if TYPE_CHECKING` blocks) and the pipeline is green.

gaborbernat (approved):

Issue #395, opened 2023-11-03
Wrap annotations with an html tag


pfebrer:
Is it possible to wrap the annotation in some custom html tag?

I want to style the whole annotation to have the same background, but looking at the generated HTML it looks like an unnecessarily daunting task, with parenthesis and commas being just text and merged with the description text. This is `Union[str, int]`:
 
![Screenshot from 2023-11-03 14-16-03](https://github.com/tox-dev/sphinx-autodoc-typehints/assets/42074085/8d5f88f0-2b11-4c3b-819a-9d36a37bfc30)

which results in background changes for `[`, `]` and `,` making it very hard to read:

![Screenshot from 2023-11-03 14-18-36](https://github.com/tox-dev/sphinx-autodoc-typehints/assets/42074085/30dfc88f-a1aa-4dc4-9445-67d148e1d103)

gaborbernat:
PR welcome.

pfebrer:
I would like to, but I don't know where to begin :sweat_smile:

pfebrer:
Should I be able to acheive this with `typehints_formatter` ?

hoodmane:
I did a similar thing in sphinx-js. It's largely a limitation of restructured text / the way that sphinx works. What I did is 
1. make a role which takes escaped rst, unescapes it, renders it into a docutils tree, and then wraps it with css span
https://github.com/pyodide/sphinx-js-fork/blob/master/sphinx_js/directives.py#L55-L66
2. when we render the type, escape your original rst and wrap it with the css anchor role:
https://github.com/pyodide/sphinx-js-fork/blob/master/sphinx_js/renderers.py#L432

I'm not sure if this is the best way, it feels a bit hacky but this was the best way I could figure out to make it work.

You can see the result here: all the types have this css wrapper. If you adjust the css for `.sphinx_js-type` you can add your box:
https://pyodide.org/en/latest/usage/api/js-api.html#pyodide

![Screenshot from 2023-11-03 11-51-48](https://github.com/tox-dev/sphinx-autodoc-typehints/assets/8739626/79d6cc77-b2e9-48a4-9799-c778b3423e3e)

pfebrer:
Thanks for the comments! 

After some investigation I realized that as you point out I have to go too deep into sphinx to implement this and I think it's not worth it for me to spend time on. I thought it would be easier, I didn't know the inner workings of sphinx were so complicated 😅

hoodmane:
[...]

PR #397, opened 2023-11-08
ENH Add a css wrapper to generated types

hoodmane:
Resolves issue #395.

Here's a screenshot where I've used this in Pyodide's docs to add some ugly css to the types:

Screenshot from 2023-11-07 16-25-51

### DIFF
--- a/src/sphinx_autodoc_typehints/__init__.py
+++ b/src/sphinx_autodoc_typehints/__init__.py
@@ -11,11 +11,13 @@ import types
 from dataclasses import dataclass
 from typing import TYPE_CHECKING, Any, AnyStr, Callable, ForwardRef, NewType, TypeVar, get_type_hints
 
+from docutils import nodes
 from docutils.frontend import OptionParser
 from docutils.parsers.rst import Parser as RstParser
+from docutils.parsers.rst import states
 from docutils.utils import new_document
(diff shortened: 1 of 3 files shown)


src/sphinx_autodoc_typehints/__init__.py
```diff
@@ -209,7 +210,7 @@ def format_annotation(annotation: Any, config: Config) -> str:  # noqa: C901, PL
     fully_qualified: bool = getattr(config, "typehints_fully_qualified", False)
     prefix = "" if fully_qualified or full_name == class_name else "~"
     role = "data" if module == "typing" and class_name in _PYDATA_ANNOTATIONS else "class"
-    args_format = "\\[{}]"
+    args_format = "\\ \\[{}]"
```
hoodmane, line 213: Maybe it would be better to use `r` here and in lots of other places: `r"\ \[{}]"` would have half as many slashes...

pfebrer:
I didn't know you could create a role like that, that's nice to know :)

gaborbernat:
@hoodmane for some reason this broke tox documentation on TypeVar:

image

https://github.com/tox-dev/tox/actions/runs/6822355429/job/18554265718

Any idea why is that?

PR #418, opened 2024-01-23
Add option to force unions (and options) to be rendered with bars

### LINKED ISSUE #413: Render union types using `|`
Python [encourages](https://github.com/python/cpython/pull/30222) using `|` over `Union` and `Optional`. s-a-t should therefore render unions that way by default.

See [here](https://discuss.python.org/t/clarification-for-pep-604-is-foo-int-none-to-replace-all-use-of-foo-optional-int/26945/6) for more context.

Of note is the argument that `Optional` is actually misleading, since it implies something that doesn’t exist in Python:
- R has true support for optional parameters via `missing(param)`, which returns `TRUE` only when the function was called without specifying `param`.
- Contrast with Python, where `param: int | None = 4` is perfectly meaningful, while the same expressed as `param: Optional[int] = 4` is very confusing: The `Optional` seems to relate to the fact that the parameter can be left out when calling the function, but actually means that you can specify `param=None` to override the default of `4`. Weird!

hoodmane:
Resolves #413.

### DIFF
--- a/src/sphinx_autodoc_typehints/__init__.py
+++ b/src/sphinx_autodoc_typehints/__init__.py
@@ -214,7 +214,7 @@ def format_annotation(annotation: Any, config: Config) -> str:  # noqa: C901, PL
     args_format = "\\[{}]"
     formatted_args: str | None = ""
 
-    always_use_bars_union: bool = getattr(config, "always_use_bars_union", False)
+    always_use_bars_union: bool = getattr(config, "always_use_bars_union", True)
     is_bars_union = full_name == "types.UnionType" or (
         always_use_bars_union and type(annotation).__qualname__ == "_UnionGenericAlias"
     )
--- a/src/sphinx_autodoc_typehints/__init__.py
+++ b/src/sphinx_autodoc_typehints/__init__.py
@@ -215,10 +215,10 @@ def format_annotation(annotation: Any, config: Config) -> str:  # noqa: C901, PL
     formatted_args: str | None = ""
 
     always_use_bars_union: bool = getattr(config, "always_use_bars_union", False)
-    isbarsunion = full_name == "types.UnionType" or (
+    is_bars_union = full_name == "types.UnionType" or (
         always_use_bars_union and type(annotation).__qualname__ == "_UnionGenericAlias"
     )
-    if isbarsunion:
+    if is_bars_union:
         full_name = ""
 
     # Some types require special handling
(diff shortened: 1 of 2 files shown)


src/sphinx_autodoc_typehints/__init__.py
```diff
@@ -214,6 +214,13 @@ def format_annotation(annotation: Any, config: Config) -> str:  # noqa: C901, PL
     args_format = "\\[{}]"
     formatted_args: str | None = ""
 
+    always_use_bars_union: bool = getattr(config, "always_use_bars_union", False)
+    isbarsunion = full_name == "types.UnionType" or (
```
gaborbernat, line 218: -> is_bars_union

src/sphinx_autodoc_typehints/__init__.py
```diff
@@ -214,6 +214,13 @@ def format_annotation(annotation: Any, config: Config) -> str:  # noqa: C901, PL
     args_format = "\\[{}]"
     formatted_args: str | None = ""
 
+    always_use_bars_union: bool = getattr(config, "always_use_bars_union", False)
```
gboeing, line 217: @hoodmane this is fantastic, thank you!

I would suggest that this should be the *only* behavior at this point in time and that the config option is unnecessary. Python itself encourages `|` over `Optional` and it therefore seems unnecessary to support an old and discouraged alternative. If the sphinx-autodoc-typehints maintainers believe it necessary to continue supporting the old way, then at a *minimum* the `always_use_bars_union` should default True. @flying-sheep any other comments?

hoodmane: > it therefore seems unnecessary to support an old and discouraged alternative. If the sphinx-autodoc-typehints maintainers believe it necessary to continue supporting the old way, then at a minimum the always_use_bars_union should default True.

I will leave this to @gaborbernat to decide. But at least this gives a way to opt into the preferred behavior, even if the discoverability isn't necessarily the best.

gaborbernat: Either is opt-in or this will need to be a major release as is in a sense backwards incompatible change.

flying-sheep: In that case, is there anything speaking against making this the default (Or not making it a setting at all and just changing the rendering)?

gaborbernat: My 2c, let's keep the flag and make it a major release. That seems to be a solution "most" would be ok with.

hoodmane: > keep the flag and make it a major release

Do you mean additionally to set the default value of the flag to `True`? Since if the default value is `False` only a minor release would be needed?

gaborbernat: I meant keep the flag, set it to True and make it a major release. The CI needs fixing too 😊

gaborbernat:
@hoodmane do you plan to pick this up again?

gaborbernat (changes requested):
CI still failing 😢

hoodmane:
Oh right. I fixed the tests but then [Make always_use_bars_union on by default](https://github.com/tox-dev/sphinx-autodoc-typehints/pull/418/commits/3b987c6b793ace47ba36f30c4f213294449733fe) broke them again.

gaborbernat (approved):

PR #422, opened 2024-01-26
Stop parsing from overwriting Sphinx configuration

### LINKED ISSUE #421: s-a-t breaks default_role in parameter descriptions
When using sphinx-autodoc-typehints, text between single backticks in parameter descriptions gets rendered as `<cite/>` nodes, i.e. using docutils’ default [`title-reference`](https://docutils.sourceforge.io/docs/ref/rst/roles.html#title-reference) role instead of the role we configured.

I don’t know exactly why, but removing s-a-t from the plugin list fixes this, and you’re e.g. using a plain `RstParser()` instance in several places which has that behavior. Maybe that parser replaces the role and doesn’t clean up properly after itself?

https://github.com/tox-dev/sphinx-autodoc-typehints/blob/0bec1bc2def5c56150fbeccf655d3fdf93c7bf12/src/sphinx_autodoc_typehints/__init__.py#L799

https://github.com/tox-dev/sphinx-autodoc-typehints/blob/0bec1bc2def5c56150fbeccf655d3fdf93c7bf12/src/sphinx_autodoc_typehints/attributes_patch.py#L66

Check out [`sphinx.testing.restructuredtext:parse`](https://github.com/sphinx-doc/sphinx/blob/master/sphinx/testing/restructuredtext.py) for a parser setup that sets up all the Sphinx stuff.

flying-sheep:
Fixes #421, fixes https://github.com/scverse/scanpy/issues/2829

This PR introduces a `parse` function that abstracts away the instantiation and use of a `RSTParser` instance.

It uses the `sphinx.util.docutils:sphinx_domains` context manager to make sure that the `docutils` directive and role registries are managed properly.

### DIFF
--- a/tests/test_sphinx_autodoc_typehints.py
+++ b/tests/test_sphinx_autodoc_typehints.py
@@ -573,7 +573,7 @@ def test_sphinx_output_default_role(app: SphinxTestApp, status: StringIO) -> Non
     set_python_path()
 
     app.config.master_doc = "simple_default_role"  # type: ignore[attr-defined] # create flag
-    app.config.default_role = "literal"
+    app.config.default_role = "literal"  # type: ignore[attr-defined]
     app.build()
 
     assert "build succeeded" in status.getvalue()  # Build succeeded
--- a/tests/test_sphinx_autodoc_typehints.py
+++ b/tests/test_sphinx_autodoc_typehints.py
@@ -554,7 +554,7 @@ def test_sphinx_output_future_annotations(app: SphinxTestApp, status: StringIO) 
        Method docstring.
 
        Parameters:
-          * **x** (bool | None) -- foo
+          * **x** (bool | None) -- *foo*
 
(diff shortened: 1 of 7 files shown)


tests/test_sphinx_autodoc_typehints.py
```diff
@@ -554,7 +554,7 @@ def test_sphinx_output_future_annotations(app: SphinxTestApp, status: StringIO)
        Method docstring.
 
        Parameters:
-          * **x** (bool | None) -- foo
+          * **x** (bool | None) -- *foo*
```
flying-sheep, line 557: The [title-reference](https://docutils.sourceforge.io/docs/ref/rst/roles.html#title-reference) role that’s used by default results in a `<cite/>` node when rendered to HTML, and apparently to `*` when rendered to text.

gaborbernat (changes requested):
Let's add a test for this case 👍

gaborbernat (approved):

PR #436, opened 2024-03-12
Skip adding pro- or epilogue when parsing

### LINKED ISSUE #425: Ignore `rst_prolog` option for function documentation
After updating from v1.24.1 to v1.25.3, the function documentation contains the `rst_prolog` option text.

The [`rst_prolog` option](https://www.sphinx-doc.org/en/master/usage/configuration.html#confval-rst_prolog) is automatically inserted at the beginning of each source file read.
I currently use it to globally limit the toc tree depth. 

With the changes between [v1.24.1 and v1.25.3](https://github.com/tox-dev/sphinx-autodoc-typehints/compare/1.24.1...1.25.3), the documentation of the individual functions is written to temporary rst files and then evaluated, as far as I understand it. This ensures that sphinx inserts the rst_prolog value and `sphinx-autodoc-typehints` seems to interpret this value not as an rst directive but as a string.

```python
# Set maximum depth for the right sidebar content overview
rst_prolog = ":tocdepth: 3"
```

Selection_20240208_01

flying-sheep:
Fixes  #425

### DIFF
--- a/src/sphinx_autodoc_typehints/parser.py
+++ b/src/sphinx_autodoc_typehints/parser.py
@@ -13,13 +13,20 @@ if TYPE_CHECKING:
 
     from docutils import nodes
     from docutils.frontend import Values
+    from docutils.statemachine import StringList
+
+
+class _RstSnippetParser(RSTParser):
+    @staticmethod
+    def decorate(_content: StringList) -> None:
+        """Skip adding pro- or epilogue."""
 
 
 def parse(inputstr: str, settings: Values | optparse.Values) -> nodes.document:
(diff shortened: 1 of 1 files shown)


gaborbernat (changes requested):
Needs a test 😊

flying-sheep:
I know. @g3n35i5 or @alexdlaird, can you please add one?

gaborbernat:
Seems stalled should we close it?

flying-sheep:
#445 took this code and runs with it so sure, as long as that one gets merged.

PR #445, opened 2024-04-09
Return injection fix

### LINKED ISSUE #444: Sphinx option rst_prolog and rst_epilog break return type injection
When adding any of those sphinx options: [rst_epilog](https://www.sphinx-doc.org/en/master/usage/configuration.html#confval-rst_epilog) or [rst_prolog](https://www.sphinx-doc.org/en/master/usage/configuration.html#confval-rst_prolog)

It will break the return type injection. For example we have one for the minimum python version (`min_python_version`). 

When using the `rst_epilog` all of the return types become the version: 
image

If you use the `rst_prolog` the return types are working again, but now the return type always appear at the top of the class, before the header and parameters.

image

Disabling either will produce the correct result:
image

Due to the rst, it creates a bunch of nodes either before or after: `<substitution_definition names="min_python_version">3.8</substitution_definition>` which mess with the line parsing.  Since it's assuming certain things come after paramters, it doesn't pick the correct lines.

caffeinepills:
Resolves #444 and #425

Ignore the rst_epilog/rst_prolog during parsing type hints like #436 did. This still keeps the replacement behavior.

I tried adjusting the parsing, but the XML node line counting, and the list of lines given are too intertwined without rewriting it completely.

`test_sphinx_autodoc_typehints` was broken for me in Windows, so I have added an encoding kwarg for it to work.

EDIT: Working on making tests pass... and adding them.

### DIFF
--- a/src/sphinx_autodoc_typehints/__init__.py
+++ b/src/sphinx_autodoc_typehints/__init__.py
@@ -779,6 +779,9 @@ def node_line_no(node: Node) -> int | None:
     docutils rst parser source code. An example where the node doesn't have a line number but the first child does is
     all `definition_list` nodes. It seems like bullet_list and option_list get line numbers, but enum_list also doesn't.
     """
+    if node is None:
+        return None
+
     while node.line is None and node.children:
         node = node.children[0]
     return node.line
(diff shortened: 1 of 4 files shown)


hoodmane:
This also needs a new test to cover the intended difference between the old behavior and the new behavior.

caffeinepills:
I added tests (might be overkill?) A single test for both epilog and prolog to make sure the replacement functionality still works. Also it will run the test config 4 times, once with the normal config, once with rst_prolog, once with rst_epilog, and once with both logs.  All tests pass.

gaborbernat (approved):

Issue #284, opened 2023-01-10
autodoc_type_aliases 


patrick-nicodemus:
As of version 3.3, Sphinx provides the functionality for a dictionary called `autodoc_type_aliases` in which the user can explicitly put type aliases that they do not want the autodoc to unravel in the documentation.

https://www.sphinx-doc.org/en/master/usage/extensions/autodoc.html#confval-autodoc_type_aliases

It would be good if this plugin would also support this feature or provide an equivalent! I was trying to understand why adding values to this dictionary was not having the desired effect as described in the documentation above and spent a while trying to debug this before I realized that this package is either interfering with that functionality or does not replace it.

This is closely related to #132, however at the current time even when using the syntax `MyType : TypeAlias = Tuple[int,str]` of https://www.python.org/dev/peps/pep-0613/ this appears not to solve the problem.

gaborbernat:
PR welcome 😊

tikuma-lsuhsc:
@patrick-nicodemus - I just pushed the PR to address this. The latest release (2.2) has it merged. So, please try it out. It should honor the `autodoc_type_aliases` config now.

hoodmane:
Closing as completed by #459.

PR #459, opened 2024-06-16
Support autodoc_type_aliases configuration

tikuma-lsuhsc:
This PR addresses Issues #284.

Currently, `autodoc-typehints` does not honor the [`autodoc_type_aliases` configuration of `sphinx.ext.autodoc`](https://www.sphinx-doc.org/en/master/usage/extensions/autodoc.html#confval-autodoc_type_aliases). For example,

```python
from numpy.typing import ArrayLike

def func(x: ArrayLike):...
```

renders

```rst
**func(x: _SupportsArray[dtype[Any]] |
_NestedSequence[_SupportsArray[dtype[Any]]] | bool | int | float |
complex | str | bytes | _NestedSequence[bool | int | float | complex |
str | bytes])
```

even if `autodoc_type_aliases = {'ArrayLike': 'ArrayLike'}` line in `conf.py`.

I modified `autodoc_typehints` to pass `app.config.autodoc_type_aliases` to `sphinx_signature()` calls to match `sphinx.ext.autodoc` and print the alias instead of the annotation as discovered.

`test_integration_autodoc_type_aliases.py` is a simple pytest script to check the conservation of `ArrayLike` annotation in a function argument.

### DIFF
--- a/tests/test_integration_autodoc_type_aliases.py
+++ b/tests/test_integration_autodoc_type_aliases.py
@@ -1,0 +1,128 @@
+from __future__ import annotations
+
+import re
+import sys
+from pathlib import Path
+from textwrap import dedent, indent
+from typing import TYPE_CHECKING, Any, Callable, NewType, TypeVar  # no type comments
+
+from numpy.typing import ArrayLike
+
+import pytest
+
+if TYPE_CHECKING:
+    from io import StringIO
+
+    from sphinx.testing.util import SphinxTestApp
+
+T = TypeVar("T")
+W = NewType("W", str)
+
+
+def expected(expected: str, **options: dict[str, Any]) -> Callable[[T], T]:
+    def dec(val: T) -> T:
+        val.EXPECTED = expected
+        val.OPTIONS = options
+        return val
+
+    return dec
+
+
+def warns(pattern: str) -> Callable[[T], T]:
+    def dec(val: T) -> T:
+        val.WARNING = pattern
+        return val
+
+    return dec
+
+
+@expected(
+    """\
+mod.function(x)
+
+   Function docstring.
+
+   Parameters:
+      **x** (ArrayLike) -- foo
+
+   Returns:
+      something
+
+   Return type:
+      bytes
+""",
+)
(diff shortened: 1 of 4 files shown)


gaborbernat (approved):

tikuma-lsuhsc:
@gaborbernat - Could you help me sort out the tox env type check errors?

I'm getting

```
tests/test_integration_autodoc_type_aliases.py:22: error: "T" has no attribute "EXPECTED"  [attr-defined]
tests/test_integration_autodoc_type_aliases.py:23: error: "T" has no attribute "OPTIONS"  [attr-defined]
tests/test_integration_autodoc_type_aliases.py:31: error: "T" has no attribute "WARNING"  [attr-defined]
tests/test_integration_autodoc_type_aliases.py:56: error: Missing return statement  [empty-body]
```

I got the last one, but not clear about the others. I created the test file by copying then stripping away existing tests of the `test_integration.py`. Thanks

gaborbernat:
Nothing jumps out for me from the top of my head. 🤔

tikuma-lsuhsc:
OK. I'll look into it when I get home tonight.

tikuma-lsuhsc:
Resolved the issue (I think).  The test file `test_integration.py` was excluded from the `mypy` type check via `pyproject.toml`. If you try to merge again, it should work now. Thanks.

tests/test_integration_autodoc_type_aliases.py
```diff
@@ -0,0 +1,125 @@
...
+
+    :param x: foo
+    :return: something
+    :rtype: bytes
+    """
+
+
+AUTO_FUNCTION = ".. autofunction:: mod.{}"
+
+LT_PY310 = sys.version_info < (3, 10)
```
gaborbernat, line 68: Inline these two to where they are used.

gaborbernat (changes requested):

tests/test_integration_autodoc_type_aliases.py
tikuma-lsuhsc: done

gaborbernat (approved):

Issue #462, opened 2024-06-19
2.2.0 Throws Exception for class with custom metaclass


b-phi:
> Handler <function process_docstring at 0x7f12ae4c8b80> for event 'autodoc-process-docstring' threw an exception (exception: unhashable type: '_SchemaMeta')

I've including the relevant skeleton of the classes below. 

```python
class _SchemaMeta(type):
    def __new__(mcs, *args: Any, **kwargs: Any) -> '_SchemaMeta':
        # some logic
        return schema_cls

    def __eq__(cls, other: Any) -> bool:
        ...

class Schema(metaclass=_SchemaMeta):
    ...
```

Elsewhere in the code, we have this which causes the error
```python
class Foo:
    def do_something(self, schema: Schema) -> None:
        """
        Args:
            schema: Some schema.
        """
        ...
```

Edit: I've pinned to 2.1.1 for now, which doesn't throw an error in this case.

hoodmane:
@tikuma-lsuhsc seems likely a regression caused by #459.

tikuma-lsuhsc:
Yep, I did not account for a string annotation as I wasn't even aware such is even an option. I'll look into it tonight.

tikuma-lsuhsc:
@b-phi - I tested all three class def's of yours but none produced the error that you reported.

Built with `sphinxcontrib-restbuilder` (for convenience as I'm using it for my current project), your code snippet produced the following without any error or warning.

```rst
**class mod._SchemaMeta(*args: Any, **kwargs: Any)**

**class mod.Schema**

**class mod.Foo**

   **do_something(mod: Schema) -> None**

      :Parameters:
         **schema** (`Schema <#mod.Schema>`_) – Some
         schema.

      :Return type:
         ``None``
```

I've also written pytest cases within the repo without any success. Could you disclose more info? (relevant `conf.py` entry & the rst lines come to my mind). Thanks.

bartfeenstra:
I am getting the following, very similar error:
```
Extension error (sphinx_autodoc_typehints):
Handler <function process_docstring at 0x7e7cf5e92520> for event 'autodoc-process-docstring' threw an exception (exception: unhashable type: 'typing.ParamSpecArgs')
```
This is also since 2.2.0 (pinning to 2.1.1 avoids the bug). I have not been able to determine the exact code sphinx-autodoc-typehints trips over. If and when I do, I'll share an example here to help reproduce the problem.

[...]

PR #463, opened 2024-06-21
Fix issue #462 and another alias processing bug

tikuma-lsuhsc:
This PR fixes the bugs introduced in v2.2:

* Error during checking for annotation alias (#462)
* Incorrect alias name look up (looking up alias name instead of the true type name)
* Aliased type ignored for the return type

The new alias look up mechanism follows closely as done in `sphinx.util.inspect.signature()`

### DIFF
--- a/src/sphinx_autodoc_typehints/__init__.py
+++ b/src/sphinx_autodoc_typehints/__init__.py
@@ -718,7 +718,11 @@ def _inject_signature(  # noqa: C901
     type_aliases = app.config["autodoc_type_aliases"]
 
     for arg_name, arg_type in signature.parameters.items():
-        annotation = arg_type.annotation if arg_type.annotation in type_aliases else type_hints.get(arg_name)
+        annotation = (
+            ForwardRef(arg_type.annotation, is_argument=True, is_class=False)
+            if arg_type.annotation in type_aliases
+            else type_hints.get(arg_name)
+        )
 
         default = signature.parameters[arg_name].default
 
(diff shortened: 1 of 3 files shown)


src/sphinx_autodoc_typehints/__init__.py
```diff
@@ -194,6 +194,9 @@ def format_annotation(annotation: Any, config: Config) -> str:  # noqa: C901, PL
     if isinstance(annotation, tuple):
         return format_internal_tuple(annotation, config)
 
+    if isinstance(annotation, TypeAliasForwardRef):
+        return str(annotation)
```
hoodmane, line 198: Should we check whether we can make an xref here? Perhaps in a followup.

hoodmane (approved):
Looks good to me if CI passes.

hoodmane (approved):
Looks good to me if CI passes.

src/sphinx_autodoc_typehints/__init__.py
tikuma-lsuhsc: sorry, I'm not following. Could you elaborate?

tikuma-lsuhsc: wait I messed up the review process. I didn't mean to dismiss your review. my bad

gaborbernat (approved):

hoodmane (approved):

hoodmane:
Thanks @tikuma-lsuhsc!

Issue #384, opened 2023-09-04
`typehints_defaults = "braces-after"` doesn't work correctly with multiline docstrings


KareemMAX:
Currently we have this [docstring](https://github.com/unifyai/ivy/blob/main/ivy/data_classes/array/activations.py#L101-L116)
```
        Parameters
        ----------
        self
            input array.
        approximate
            whether to use the approximate version of the gelu function.
        complex_mode
            optional specifier for how to handle complex data types. See
            ``ivy.func_wrapper.handle_complex_input`` for more detail.
        out
            optional output array, for writing the result to. It must have a shape
            that the inputs broadcast to.
```

When we use `typehints_defaults = "braces-after"` it renders incorrectly for `complex_mode`:
<img width="698" alt="image" src="https://github.com/tox-dev/sphinx-autodoc-typehints/assets/16426366/60811a2d-3bd9-40cc-bdfe-9d2240ff00a5">

The `(default: 'jax')` should be places at the end of line 2 (after "more details."), while it's currently places after line 1 (See).

gaborbernat:
PR welcome 👍

PR #464, opened 2024-06-22
Fix for Issue #384: typehints_defaults = "braces-after" fails for a multiline `:param:` entry

tikuma-lsuhsc:
Here is another quick fix.

The fix advances the line to the last line of the last `:param:` paragraph by detecting either that the next line is empty or has no indentation and placing the default at the end of the last nonempty line.

I've read somewhere else that rst docstring spec does not explicitly mention anything about multiline argument descriptions, but I understand it is widely accepted to use hanging indentation.

@gaborbernat & @hoodmane - One thing I'm not sure of is the handling of a multi-paragraph `:param:` description. It is likely rare to document in such a way. But when it does, should the default be placed at the end of the first or the last paragraph? 

I've gone with the latter for now, but I can see myself preferring the default to appear on the first paragraph when it happens. (2nd+ paragraphs for in-depth descriptions.) Ultimately, we can add a new option for it, but a default behavior must be set regardless. The current on-the-last-paragraph implementation is simply the last one I thought of, and I can go either way. Let me know what you guys think, and I'd be happy to modify to suit your preference.

### DIFF
--- a/src/sphinx_autodoc_typehints/__init__.py
+++ b/src/sphinx_autodoc_typehints/__init__.py
@@ -741,11 +741,6 @@ def _inject_signature(  # noqa: C901
 
         if insert_index is not None:
 
-            # advance index to the end of paragraph
-            # nlines = len(lines)
-            # while insert_index < nlines and (lines[insert_index + 1] == "" or lines[insert_index + 1].startswith(" ")):
-            #     insert_index += 1
-
             if annotation is None:
                 type_annotation = f":type {arg_name}: "
             else:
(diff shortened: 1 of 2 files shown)


gaborbernat:
CI broken 😎

tikuma-lsuhsc:
@gaborbernat - Working on it... too many nesting? y'all are picky!/s

src/sphinx_autodoc_typehints/__init__.py
```diff
@@ -749,14 +749,33 @@ def _inject_signature(  # noqa: C901
...
 
+def _append_default(
+    app: Sphinx, lines: list[str], insert_index: int, type_annotation: str, formatted_default: str
+) -> str:
+    if app.config.typehints_defaults.endswith("after"):
+        # advance the index to the end of the :param: paragraphs
+        # (terminated by a line with no indentation)
+        # append default to the last nonempty line
+        nlines = len(lines)
+        i = insert_index + 1
```
gaborbernat, line 765: The only place I'm okay using single letter variable is within list compressions. Anywhere else single letter variables are so abstract that they are really hard to read. Please use more descriptive names.

gaborbernat (changes requested):

src/sphinx_autodoc_typehints/__init__.py
tikuma-lsuhsc: done

gaborbernat (approved):

gaborbernat (approved):

PR #468, opened 2024-07-17
Fix rendering when docstring starst with a directive

### LINKED ISSUE #467: Conflict with directives `seealso`, `deprecated`, ...
Minimal reproduction example:

```python

def hello() -> str:
    """
    .. seealso:: more info at <https://example.com>`_.
    """
    return "Hello, world!"
```

This triggers this warning during build:

> .../helloworld.py:docstring of helloworld.hello:3: WARNING: Explicit markup ends without a blank line; unexpected unindent.

(using sphinx==7.4.1, sphinx_autodoc_typehints==2.2.2)

hoodmane:
We want to position our insertion before the directive but we incorrectly get a negative insertion line. This inserts at the end rather than at the beginning, and we fail to leave the correct empty line after the directive.

Resolves #467


gaborbernat (changes requested):
We just need to fix the CI 👍

hoodmane:
Opened #469 since the failure is on the main branch.

gaborbernat (approved):

PR #487, opened 2024-09-20
Add typehints_rtype_after_returns option

Kevin-Roberts:
* Allow the user to specify the order of the rtype/returns statements.
* Fix an index error exception caused when always_document_param_types is True and typehints_defaults is 'braces-after'.

### DIFF
--- a/src/sphinx_autodoc_typehints/__init__.py
+++ b/src/sphinx_autodoc_typehints/__init__.py
@@ -778,8 +778,11 @@ def _append_default(
             if lines[next_index]:
                 append_index = next_index
             next_index += 1
-        lines[append_index] += formatted_default
-
+        if append_index >= nlines:
+            # When append_index is greater than nlines, simply append the line to the end of lines
+            lines.append(formatted_default)
+        else:
+            lines[append_index] += formatted_default
     else:  # add to last param doc line
         type_annotation += formatted_default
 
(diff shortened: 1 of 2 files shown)


Kevin-Roberts:
So the bug is easy to recreate, just set always_document_param_types=True, typehints_document_rtype=True, typehints_use_rtype=True, typehints_defaults='before-after' (this parameter that ends with *-after is what causes the bug) and have a function with a simple docstring with a description and nothing else, for example:

def my_test(variable_1: str) -> str:
   """ 
   bla bla
   """

src/sphinx_autodoc_typehints/__init__.py
```diff
@@ -778,8 +778,11 @@ def _append_default(
             if lines[next_index]:
                 append_index = next_index
             next_index += 1
-        lines[append_index] += formatted_default
-
+        if append_index >= nlines:
```
Kevin-Roberts, line 781: Without this you get an index error in some cases (see my general comment)

src/sphinx_autodoc_typehints/__init__.py
```diff
@@ -895,6 +898,9 @@ def _inject_rtype(  # noqa: PLR0913, PLR0917
 
     formatted_annotation = add_type_css_class(format_annotation(type_hints["return"], app.config))
 
+    if app.config.typehints_rtype_after_returns and r.found_return:
```
Kevin-Roberts, line 901: I think general sphinx users often put the rtype after returns, so adding the ability to choose.

gaborbernat (changes requested):
Please notify me when the CI passes and you have added tests.

Kevin-Roberts:
Sounds good.. You see what I am going for though right

hoodmane:
Can you explain why a user would care about this setting? What difference does it make in the rendered output?

Kevin-Roberts:
Well so I'm trying to use this plugin for a fairly large project and right now we always have :rtype: after :returns: and I'm trying to avoid having every functions rendered output change (mostly due to automated emails). 

Then there is also the bug fix when combining current config options which 100% should be fixed.

Issue #457, opened 2024-05-28
Sphinx raises warning about custom typehints_formatter


jl-wynen:
Since Sphinx v7.3.0, there is a warning about config values that are not pickleable. When building with `warning_is_error = True`, this fails the build. See https://github.com/sphinx-doc/sphinx/issues/12300 for context.

For sphinx-autodoc-typehints, this means that a custom `typehints_formatter` leads to a build failure.

To reproduce, simply use this `conf.py`:
```python
project = 'custom_formatter'
extensions = [
    'sphinx.ext.autodoc',
    'sphinx_autodoc_typehints',
]
warning_is_error = True
show_warning_types = True

def formatter(annotation, config):
    return None

typehints_formatter = formatter
```
It results in this warning:
```
sphinx.errors.SphinxWarning: cannot cache unpickable configuration value: 'typehints_formatter' (because it contains a function, class, or module object) [config.cache]
```

Apart from triggering a warning that needs to be silenced, this also prevents caching. So it would be great to have a solution that avoids specifying a callable as `typehints_formatter`. One option would be using a fully qualified path and make sphinx-autodoc-typehints import a function from that path. E.g. `typehints_formatter = 'mypackage.sphinxext.typehints_formatter'`. See also this comment: https://github.com/sphinx-doc/sphinx/issues/12300#issuecomment-2061054462

bckohan:
I've had this same issue in [sphinxcontrib/typer](https://github.com/sphinx-contrib/typer). 

The only way to fix it is the method you suggest, where you load it from a string import path.

gaborbernat:
Validated that this no longer reproduces - setting `typehints_formatter` to a callable no longer triggers `cannot cache unpickable configuration value` warnings. The project now requires Sphinx >=9.1 which handles this differently.

If you still encounter this, please open a new issue with a reproducer.

PR #494, opened 2024-10-08
Apply typehints_formatter to signature

### LINKED ISSUE #429: typehints_formatter not applied to signature
When type hints are shown in the signature, they are not passed through `typehints_formatter`.

## Example

```
package
├── docs
│   ├── conf.py
│   └── index.rst
├── my_module
│   └── __init__.py
└── pyproject.toml
```

conf.py
```python
extensions = [
    "sphinx.ext.autodoc",
    "sphinx.ext.napoleon",
    "sphinx_autodoc_typehints",
    "sphinx.ext.autosummary",
]

# sphinx.ext.napoleon
napoleon_use_rtype = typehints_use_rtype = False

# sphinx.ext.autodoc
# For this issue, this makes no difference regarding the signature, only regarding the description.
# Note that "description" and "both" add a separate "Return type" section which is also not
# formatted by typehints_formatter. 
autodoc_typehints = "both"

# sphinx.ext.autosummary
autosummary_generate = True

# sphinx_autodoc_typehints
typehints_document_rtype = False
typehints_use_signature = True
typehints_use_signature_return = True


def typehints_formatter(annotation, config):
    from typing import TypeVar

    if isinstance(annotation, TypeVar) and annotation.__name__ == "Unformatted":
        return "Formatted"

```

index.rst
```reStructuredText
.. autosummary::

   my_module.my_function
```

my\_module/\_\_init\_\_.py
```python
from typing import TypeVar

Unformatted = TypeVar("Unformatted")


def my_function(arg: Unformatted) -> Unformatted:
    """
    Do nothing

    Args:
        arg: An argument

    Returns:
        The argument returned
    """
    return arg

```

## Expected behavior

- Types are formatted everywhere:
  - Parameter types in signature (because `typehints_use_signature = True`)
  - Return type in signature (because `typehints_use_signature_return = True`)
  - Types in "Parameters" section of description (because `autodoc_typehints = "both"`)
  - Types in "Returns" section of description (because `autodoc_typehints = "both"`)
  - Types in "Return type" section of description, if shown (if `napoleon_use_rtype = typehints_use_rtype = True`)

## Actual behavior

- Parameter types in signature are not formatted.
- Return type in signature is not formatted.

image

## Versions

```
Platform:              linux; (Linux-5.15.0-91-generic-x86_64-with-glibc2.35)
Python version:        3.9.17 (main, Jul  5 2023, 20:41:20) 
[GCC 11.2.0])
Python implementation: CPython
Sphinx version:        7.2.6
Docutils version:      0.20.1
Jinja2 version:        3.1.2
Pygments version:      2.16.1
sphinx-autodoc-typehints: 2.0.0
```

Priyansh121096:
Fixes #429

### DIFF
--- a/tests/test_integration.py
+++ b/tests/test_integration.py
@@ -1406,7 +1406,7 @@ Unformatted = TypeVar("Unformatted")
     """,
     typehints_use_signature=True,
     typehints_use_signature_return=True,
-    typehints_formatter=lambda ann, cfg=None: "Formatted",
+    typehints_formatter=lambda _, __=None: "Formatted",
 )
 def typehints_formatter_applied_to_signature(param: Unformatted) -> Unformatted:
     """
--- a/tests/test_integration.py
+++ b/tests/test_integration.py
@@ -1385,6 +1385,41 @@ def has_doctest1() -> None:
     """
 
 
+Unformatted = TypeVar("Unformatted")
+
+@warns("cannot cache unpickable configuration value: 'typehints_formatter'")
+@expected(
+    """
+    mod.typehints_formatter_applied_to_signature(param: Formatted) -> Formatted
+
+       Do nothing
+
+       Parameters:
+          **param** (Formatted) -- A parameter
+
+       Return type:
+          Formatted
+
+       Returns:
+          The return value
+    """,
+    typehints_use_signature=True,
+    typehints_use_signature_return=True,
+    typehints_formatter=lambda ann, cfg=None: "Formatted",
+)
+def typehints_formatter_applied_to_signature(param: Unformatted) -> Unformatted:
+    """
+    Do nothing
+
+    Args:
+        param: A parameter
+
+    Returns:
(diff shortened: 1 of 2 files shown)


tests/test_integration.py
```diff
@@ -1385,6 +1385,42 @@ def has_doctest1() -> None:
     """
 
 
+Unformatted = TypeVar("Unformatted")
+
+
+@warns("cannot cache unpickable configuration value: 'typehints_formatter'")
```
gaborbernat, line 1391: Feels like we should fix this warning not ignore it 🤔

gaborbernat (changes requested):

tests/test_integration.py
Priyansh121096: #457 is an open issue for this.

gaborbernat (approved):

PR #517, opened 2025-01-16
Update Ruff to 0.9.1

flying-sheep:
This needs a few manual changes

### DIFF
--- a/tests/test_integration.py
+++ b/tests/test_integration.py
@@ -148,10 +148,10 @@ class Class:
     :param z: baz
     """
 
-    def __init__(self, x: bool, y: int, z: Optional[str] = None) -> None:  # noqa: UP045
+    def __init__(self, x: bool, y: int, z: Optional[str] = None) -> None:
         pass
 
-    def a_method(self, x: bool, y: int, z: Optional[str] = None) -> str:  # noqa: UP045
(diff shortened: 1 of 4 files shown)


tests/test_integration.py
```diff
@@ -148,10 +148,10 @@ class Class:
     :param z: baz
     """
 
-    def __init__(self, x: bool, y: int, z: Optional[str] = None) -> None:  # noqa: UP007
+    def __init__(self, x: bool, y: int, z: Optional[str] = None) -> None:  # noqa: UP045
```
hoodmane, line 151: Can we disable `UP045` for the entire tests directory?

flying-sheep: If you want to do that, you can do

```toml
[tool.ruff.lint.per-file-ignores]
"tests/*" = ["UP45"]
```

hoodmane: Could you add that then I'll merge this?

flying-sheep: Sure, I also added UP006 and UP007 since that’s the same thing

hoodmane (approved):
Thanks for helping with the chores @flying-sheep!

flying-sheep:
No problem! Putting on the software janitor hat is as rewarding as being a real janitor, but often much less gross!

Issue #523, opened 2025-02-17
_stringify_annotation patch breaks in Sphinx==8.2.0


b-kamphorst:
Hi!

I just tested `sphinx-autodoc-typehints==3.0.1` against the release candidate `sphinx==8.2.0rc1` and it breaks. Reason is that [the signature of `stringify_annotation`](https://github.com/sphinx-doc/sphinx/blob/v8.2.0rc1/sphinx/util/typing.py#L393) got an additional (keyword-only) parameter `short_literals`, which is not accepted by `sphinx-autodoc-typehints` [patched](https://github.com/tox-dev/sphinx-autodoc-typehints/blob/3.0.1/src/sphinx_autodoc_typehints/attributes_patch.py#L57) version named [`_stringify_annotation`](https://github.com/tox-dev/sphinx-autodoc-typehints/blob/3.0.1/src/sphinx_autodoc_typehints/attributes_patch.py#L45).

An easy but inelegant fix would be to accept and ignore any `*args` and `**kwargs`, but I'm not sufficiently familiar with this repo to know whether you actually need to do something with the new argument.

<details><summary>Full traceback</summary>

```console
Traceback
=========

    Traceback (most recent call last):
      File "/usr/local/lib/python3.12/site-packages/sphinx/cmd/build.py", line 432, in build_main
        app.build(args.force_all, args.filenames)
      File "/usr/local/lib/python3.12/site-packages/sphinx/application.py", line 426, in build
        self.builder.build_update()
      File "/usr/local/lib/python3.12/site-packages/sphinx/builders/__init__.py", line 375, in build_update
        self.build(
      File "/usr/local/lib/python3.12/site-packages/sphinx/builders/__init__.py", line 403, in build
        updated_docnames = set(self.read())
                               ^^^^^^^^^^^
      File "/usr/local/lib/python3.12/site-packages/sphinx/builders/__init__.py", line 519, in read
        self._read_serial(docnames)
      File "/usr/local/lib/python3.12/site-packages/sphinx/builders/__init__.py", line 584, in _read_serial
        self.read_doc(docname)
      File "/usr/local/lib/python3.12/site-packages/sphinx/builders/__init__.py", line 648, in read_doc
        publisher.publish()
      File "/usr/local/lib/python3.12/site-packages/docutils/core.py", line 234, in publish
        self.document = self.reader.read(self.source, self.parser,
[...]

PR #525, opened 2025-02-19
Support Sphinx 8.2.0 - drop 3.10 support because Sphinx does

b-kamphorst:
Closes https://github.com/tox-dev/sphinx-autodoc-typehints/issues/523.


b-kamphorst:
This PR fixes some incompatibilities with sphinx 8.2.0. Most notable changes:
- mock `_stringify_annotation` now accepts arbitrary `*args` and `**kwargs`, which may or may not be desired
- adds logic for dealing with `python_display_short_literal_types` (which is the reason the mock broke)

Please take additional care in those changes as they are opinionated / change behaviour. Especially the second one: revisit the tests and ensure that the short syntax produces valid and expected rst.

gaborbernat (approved):

Issue #481, opened 2024-09-11
Missing links for a union with an unevaluated type alias


nineteendo:
It's not a big deal, but I would like the links to the type definitions to be present on all pages: https://jsonyx.readthedocs.io/en/latest/api/jsonyx.apply_patch.html

```python
# conf.py
autodoc_type_aliases = {"_Operation": "_Operation"}
```

```python
_Operation = dict[str, Any]

def apply_patch(
    obj: Any,
    patch: _Operation | list[_Operation],
    *,
    allow: Container[str] = NOTHING,
    use_decimal: bool = False,
) -> Any:
    ...
```

<img width="653" alt="Screenshot 2024-09-11 at 20 23 12" src="https://github.com/user-attachments/assets/a8555bef-9c03-41c6-815a-50c82812bf53">

nineteendo:
`typing.get_type_hints()` raises this error: 

```python
TypeError("unsupported operand type(s) for |: 'TypeAliasForwardRef' and 'types.GenericAlias'")
```

This seems to fix the issue:

```diff
-localns = TypeAliasNamespace(app.config["autodoc_type_aliases"])
+localns = {key: ForwardRef(value) for key, value in app.config["autodoc_type_aliases"].items()}
```

PR #533, opened 2025-03-30
Fix issue #481

nineteendo:


hoodmane:
Please add a test.

nineteendo:
What should I do with `"Class Alias"`? That's not a valid forward reference.

hoodmane:
The documentation of autodoc-type-aliases says:

> A dictionary for users defined [type aliases](https://mypy.readthedocs.io/en/latest/kinds_of_types.html#type-aliases) that maps a type name to the full-qualified object name. It is used to keep type aliases not evaluated in the document.

https://www.sphinx-doc.org/en/master/usage/extensions/autodoc.html#confval-autodoc_type_aliases

`Class Alias` isn't a fully qualified object name or even an identifier.

hoodmane:
But I don't understand the problem or the fix very well @nineteendo could you explain a bit more?

nineteendo:
1. Aliases are converted to `sphinx.util.inspect.TypeAliasForwardRef`
2. `Alias | A` is not supported.
3. This crashes `typing.get_type_hints()`
4. All annotations become strings.

I fixed this by creating a subclass that defines `__or__()`.

src/sphinx_autodoc_typehints/__init__.py
```diff
@@ -689,7 +696,7 @@ def process_docstring(  # noqa: PLR0913, PLR0917
     except (ValueError, TypeError):
         signature = None
 
-    localns = TypeAliasNamespace(app.config["autodoc_type_aliases"])
```
hoodmane, line 692: `TypeAliasNamepace` seems to have a fair amount of logic that this removes. We apparently don't have any coverage for it. But perhaps it would be better to monkeypatch `__or__` onto `TypeAliasForwardRef`?

nineteendo: It just gets passed to `typing.get_type_hints()`, which gets converted to a dict and then passed to `eval()`.
And I'm not a huge fan of monkeypatching.

hoodmane: What about these lines: https://github.com/sphinx-doc/sphinx/blob/master/sphinx/util/inspect.py?plain=1#L670-L676

nineteendo: They are for convenience: `ns.typing.Any` instead of `ns["typing.Any"]`, but modules are invalid type hints:

```python
>>> from sphinx.util.inspect import TypeAliasNamespace
>>> ns = TypeAliasNamespace({"typing.Any": "Any"})
>>> dict(ns)
{
    'typing.Any': TypeAliasForwardRef('Any'),
    'typing': <sphinx.util.inspect.TypeAliasModule object at 0x000001B1A33C4A50>
}
```

nineteendo: This is what happens if you try:

`<unknown>:1: WARNING: py:class reference target not found: sphinx.util.inspect.TypeAliasModule [ref.class]`

image

hoodmane: Okay, I still don't fully understand but if `TypeAliasNamepace` has no benefit for us / causes trouble then I think you're approach makes sense.

nineteendo:
Hmm, forward references are not subscriptable either:

```none
WARNING: Cannot resolve forward reference in type annotations of "jsonyx.dump": 'MyTypeAliasForwardRef' object is
not subscriptable
```

nineteendo:
Here a potential implementation for that (I've had enough of it):

```diff
diff --git a/src/sphinx_autodoc_typehints/__init__.py b/src/sphinx_autodoc_typehints/__init__.py
index 4f20b97..0231a91 100644
--- a/src/sphinx_autodoc_typehints/__init__.py
+++ b/src/sphinx_autodoc_typehints/__init__.py
@@ -65,6 +65,9 @@ class MyTypeAliasForwardRef(TypeAliasForwardRef):
     def __or__(self, value: Any) -> Any:
         return Union[self, value]  # noqa: UP007
 
+    def __getitem__(self, item: Any) -> types.GenericAlias:
+        return types.GenericAlias(self, item)  # type: ignore
+
 
 def _get_types_type(obj: Any) -> str | None:
     try:
@@ -227,9 +230,6 @@ def format_annotation(annotation: Any, config: Config, *, short_literals: bool =
     if isinstance(annotation, tuple):
         return format_internal_tuple(annotation, config)
 
-    if isinstance(annotation, TypeAliasForwardRef):
-        return annotation.name
-
     try:
         module = get_annotation_module(annotation)
         class_name = get_annotation_class_name(annotation, module)
@@ -300,6 +300,10 @@ def format_annotation(annotation: Any, config: Config, *, short_literals: bool =
         formatted_args = args_format.format(", ".join(fmt))
 
     escape = "\\ " if formatted_args else ""
+    annotation = getattr(annotation, "__origin__", annotation)
+    if isinstance(annotation, TypeAliasForwardRef):
+        return f"{annotation.name}{escape}{formatted_args}"
+
     return f":py:{role}:`{prefix}{full_name}`{escape}{formatted_args}"
```

gaborbernat (approved):

PR #576, opened 2025-10-15
Preserve type information from docstrings if no type annotation is present and parameter has default value.

### LINKED ISSUE #574: Parameters with a default value lose their type information from docstrings
Consider the following function, which features a parameter that has a default value:

```python
def foobar(name, surname='Smith'):
    """
    This is an example.

    :type name: str
    :type surname: str
    :param name: The name of the user.
    :param surname: The surname of the user.
    """
```

If **sphinx-autodoc-typehints** is configured with

```python
typehints_defaults = 'comma'
typehints_use_signature = True
```

, the mere existence of the default value makes the extension disregard the type information that is present in the docstring:

Current behavior: Type information missing

_Ideally_, if the parameter provides a default value but not a type hint, the type information in the docstring should be used if it is present:

Expected behavior: Correct type information

> [!IMPORTANT]
> The issue is reproducible if `typehints_defaults` is set to `'comma'` or `'braces'`, but not if set to `'braces-after'`.

christianaguilera-foundry:
Fixes #574.

### DIFF
--- a/src/sphinx_autodoc_typehints/__init__.py
+++ b/src/sphinx_autodoc_typehints/__init__.py
@@ -808,43 +808,63 @@ def _inject_signature(
     lines: list[str],
 ) -> None:
     for arg_name in signature.parameters:
-        annotation = type_hints.get(arg_name)
+        _inject_arg_signature(type_hints, signature, app, lines, arg_name)
 
-        default = signature.parameters[arg_name].default
 
-        if arg_name.endswith("_"):
-            arg_name = f"{arg_name[:-1]}\\_"  # noqa: PLW2901
+def _inject_arg_signature(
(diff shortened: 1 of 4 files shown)


gaborbernat:
No th ones added are fine.

christianaguilera-foundry:
I think I got the unit tests right now.

For the records, I had some weird issues where, at some point, the unit tests would fail with some odd stale messages (switching to `main`; without my changes; would still fail). Then I had issues with "coverage" (going from 88% to 84%),  even though the code I added is has 100% coverage (which means the overall coverage would increase). What I did to fix is was to remove the `.tox` directory and re-run; unsure how much of this was human error, or some known gotchas.

Issue #580, opened 2025-10-27
“Cannot resolve forward reference” when using a type variable defined by Python 3.12 type syntax


flying-sheep:
```python
from __future__ import annotations

class Foo[T]:
   def __init__(self, thing: T) -> None: pass
```

will result in `sphinx.errors.SphinxWarning: Cannot resolve forward reference in type annotations of "mylib.Foo": name 'T' is not defined`

This should not be the case, as the `[T]` syntax defines it. See [type parameter lists](https://docs.python.org/3/reference/compound_stmts.html#type-params)/
[PEP 696](https://peps.python.org/pep-0696/)

flying-sheep:
Still there with 3.6.2.

IDK why the traceback is cut off when you run this with `--excepion-on-warning`

```pytb
Traceback (most recent call last):
  File "…/site-packages/sphinx_autodoc_typehints/__init__.py", line 554, in _get_type_hint
    result = get_type_hints(obj, None, localns, include_extras=True)
  File "/usr/lib/python3.14/typing.py", line 2437, in get_type_hints
    value = _eval_type(value, globalns, localns, (), format=format, owner=obj, prefer_fwd_module=True)
  File "/usr/lib/python3.14/typing.py", line 461, in _eval_type
    return evaluate_forward_ref(t, globals=globalns, locals=localns,
                                type_params=type_params, owner=owner,
                                _recursive_guard=recursive_guard, format=format)
  File "/usr/lib/python3.14/typing.py", line 1000, in evaluate_forward_ref
    value = forward_ref.evaluate(globals=globals, locals=locals,
                                 type_params=type_params, owner=owner, format=format)
  File "/usr/lib/python3.14/annotationlib.py", line 201, in evaluate
    raise NameError(_NAME_ERROR_MSG.format(name=arg), name=arg)
NameError: name 'T' is not defined
```

flying-sheep:
This is https://github.com/python/cpython/issues/124089

PR #595, opened 2026-01-02
Fix compatibility with 9.1.0

gaborbernat:
Signed-off-by: Bernát Gábor <bgabor8@bloomberg.net>


flying-sheep:
That didn’t do the trick yet, I still see #580 with 3.6.2

PR #605, opened 2026-02-17
Fix IndexError with always_document_param_types and braces-after defaults

### LINKED ISSUE #575: `IndexError` may be raised if `typehints_defaults = 'braces-after'` is combined with `always_document_param_types = True`
The exception is:
```
    Traceback (most recent call last):
      File ".../sphinx/events.py", line 404, in emit
        results.append(listener.handler(self.app, *args))
                       ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
      File ".../sphinx_autodoc_typehints/__init__.py", line 743, in process_docstring
        _inject_types_to_docstring(type_hints, signature, original_obj, app, what, name, lines)
      File ".../sphinx_autodoc_typehints/__init__.py", line 799, in _inject_types_to_docstring
        _inject_signature(type_hints, signature, app, lines)
      File ".../sphinx_autodoc_typehints/__init__.py", line 845, in _inject_signature
        type_annotation = _append_default(app, lines, insert_index, type_annotation, formatted_default)
                          ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
      File ".../sphinx_autodoc_typehints/__init__.py", line 864, in _append_default
        lines[append_index] += formatted_default
        ~~~~~^^^^^^^^^^^^^^
    IndexError: list index out of range
```
This code path is only reached if `typehints_defaults` is set to `'braces-after'`.

It seems that, if `always_document_param_types` is enabled, `insert_index` may take a value that is out of bounds:

```python
        if annotation is not None and insert_index is None and app.config.always_document_param_types:
            lines.append(f":param {arg_name}:")
            insert_index = len(lines)
```

A naive solution would be to subtract 1 so that `insert_index` stores the index to the last line. EDIT: This suggestion may or may not be accurate.

worksbyfriday:
Fixes #575

When `always_document_param_types` is `True` and a parameter is not documented in the docstring, `_inject_arg_signature` appends a new `:param:` line and sets `insert_index` to `len(lines)`. This is one past the last valid index, causing an `IndexError` in `_append_default` when `typehints_defaults` is `'braces-after'`.

The fix: set `insert_index` to `len(lines) - 1` (the index of the just-appended line).

Includes a regression test that reproduces the exact crash from the issue.

### DIFF
--- a/tests/test_sphinx_autodoc_typehints.py
+++ b/tests/test_sphinx_autodoc_typehints.py
@@ -656,13 +656,10 @@ def test_always_document_param_types_with_defaults_braces_after(
 
     for rst_file in Path(app.srcdir).glob("*.rst"):
         rst_file.unlink()
-    (Path(app.srcdir) / "index.rst").write_text(
-        dedent(
-            """
-            .. autofunction:: dummy_module.undocumented_function_with_defaults
-            """,
-        ),
-    )
+    index_content = """\
+        .. autofunction:: dummy_module.undocumented_function_with_defaults
+    """
+    (Path(app.srcdir) / "index.rst").write_text(dedent(index_content))
 
     app.build()
 


Fridayai700:
Fixed the pre-commit check — added `# noqa: ARG001` to the `warning` fixture parameter (it's a pytest fixture, can't be renamed). All checks passing now.

tests/test_sphinx_autodoc_typehints.py
```diff
@@ -587,6 +587,34 @@ class dummy_module.DataClass(x)
...
+    status: StringIO,
+    warning: StringIO,  # noqa: ARG001
+) -> None:
+    """Regression test for #575: IndexError when combining always_document_param_types with braces-after."""
+    set_python_path()
+
+    app.config.always_document_param_types = True
+    app.config.typehints_defaults = "braces-after"
+
+    for f in Path(app.srcdir).glob("*.rst"):
```
gaborbernat: don't use single letter variables outside of list comprehensions.

gaborbernat (changes requested):

Fridayai700:
Moved `dedent()` into the `write_text()` call to match the existing pattern and eliminate the intermediate variable.

gaborbernat (approved):

PR #297, opened 2023-01-17
Add fix for #296

hoodmane:
See the explanation in the diff and in issue #296.

PR #625, opened 2026-02-23
✨ feat(signatures): enable overload signature display

### LINKED ISSUE #529: Support overloads
#297 did the opposite of what the author of any moderately complex overloaded signature wants: it removes it from the docs.

We should make sure all overloaded signatures appear in the docs, with their types rendered by s-a-t.

gaborbernat:
PR #297 removed overload signature support to work around Sphinx not emitting `autodoc-process-signature` events for overloaded functions. This solved a technical issue but removed valuable type information from documentation. Users with complex overloaded APIs lost their distinct signature variants, making it harder to understand which types are valid for different use cases. 📚

The solution renders overloads in the docstring description instead of in signature lines. Overloads are cached during module analysis, then injected into docstrings as a formatted field list with proper RST cross-references. This avoids Sphinx's architectural limitation (signatures can't contain RST markup) while giving users complete type information. ✨

Functions decorated with `@overload` now show:

```
Overloads:
  * x (int) → str
  * x (list[int]) → dict[str, int]
```

All types are clickable cross-references in HTML output. Generic return types are automatically suppressed when overloads are present since each variant shows its specific return type.

Resolves #529

### DIFF
--- a/src/sphinx_autodoc_typehints/__init__.py
+++ b/src/sphinx_autodoc_typehints/__init__.py
@@ -808,7 +808,11 @@ def process_docstring(  # noqa: PLR0913, PLR0917
 
 
 def _inject_overload_signatures(  # noqa: C901
-    app: Sphinx, what: str, name: str, obj: Any, lines: list[str]  # noqa: ARG001
+    app: Sphinx,
+    what: str,
+    name: str,
+    obj: Any,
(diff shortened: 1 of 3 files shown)


gaborbernat:
pre-commit.ci autofix

PR #671, opened 2026-03-24
🐛 fix(stubs): evaluate stub-only definitions for annotation resolution

gaborbernat:
Stub files define names that don't exist at runtime: `TypeVar` definitions like `_ScalarT = TypeVar("_ScalarT", bound=generic)` and `TypedDict` classes decorated with `@type_check_only`. These stub-only names were missing from the evaluation namespace, causing `NameError` warnings when `get_type_hints` tried to resolve annotations referencing them.

This extends `_resolve_stub_imports` to also evaluate top-level definitions (assignments, classes, type aliases) from stubs after imports are resolved. Class decorators like `@type_check_only` are stripped before execution since they may not exist at runtime.

Validated against the numpy reproduction from #669 (`_ScalarT`/`_UFuncKwargs` warnings eliminated) and confirmed no regressions against cbor2 and matplotlib test cases.

Closes #669

### DIFF
--- a/src/sphinx_autodoc_typehints/_resolver/_stubs.py
+++ b/src/sphinx_autodoc_typehints/_resolver/_stubs.py
@@ -7,6 +7,7 @@ import contextlib
 import importlib
 import inspect
 import sys
+from copy import copy
 from pathlib import Path
 from typing import TYPE_CHECKING, Any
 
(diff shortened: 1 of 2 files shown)


thomasWeise:
Thank you a thousand times.
Now my whole build chain runs smooth without any warning or error.

PR #691, opened 2026-04-30
Support PEP 695 `type` statement and python 3.12+ `TypeAliasType`

### LINKED ISSUE #690: `type X = Y` (PEP 695 `TypeAliasType`) renders as `TypeAliasType` instead of alias value
## Environment
sphinx: 9.1.0
sphinx-autodoc-typehints version: 3.10.2
Python version: 3.12+

## Problem
`typing.TypeAliasType` instances created by the PEP 695 [`type` statement](https://docs.python.org/3/reference/simple_stmts.html#type) or `TypeAliasType(...)` directly are  not recognised by `format_annotation`. Instead of resolving the alias or emitting a cross-reference, the formatter falls through to the generic class path and renders the metaclass:
```
:py:class:`typing.TypeAliasType`
```
This affects both **locally-defined** and **external** aliases resolved via intersphinx. 

### Local alias
The [`load_dataset`](https://github.com/neuroinformatics-unit/movement/blob/8f2a315c1bfac3b383bc923f3c7a69bb6d1d1c98/movement/io/load.py#L321-L326) function has `source_software` annotated with `SourceSoftwareOrAuto`, a `TypeAliasType`. 
```python
type SourceSoftware = Literal[
    "DeepLabCut",
    "SLEAP",
    "LightningPose",
    "Anipose",
    "NWB",
    "VIA-tracks",
]
type SourceSoftwareOrAuto = (
    SourceSoftware | Literal["auto", "DeepLabCut/LightningPose"]
)

def load_dataset(
    file: Path | str | pynwb.file.NWBFile,
    source_software: SourceSoftwareOrAuto = "auto",
    fps: float | None = None,
    **kwargs,
) -> xarray.Dataset:
    """Create a ``movement`` dataset from any supported third-party file. """
```
| Actual | Expected |
| ------- | ---------- |
| Image | Image |

### External alias (intersphinx)
The [`scale`](https://github.com/neuroinformatics-unit/movement/blob/8f2a315c1bfac3b383bc923f3c7a69bb6d1d1c98/movement/transforms.py#L14-L20) function has `factor` annotated with a Union of `numpy.typing.ArrayLike` (a `TypeAliasType` in numpy) and `float`. 

```python
@log_to_attrs
def scale(
    data: xarray.DataArray,
    factor: numpy.typing.ArrayLike | float,
    space_unit: str | None = None,
) -> xarray.DataArray:
    """Scale data by a given factor with an optional unit. """
```
| Actual | Expected |
| ------- | ---------- |
| Image | Image |

In both cases, since `format_annotation` does not have a branch for `TypeAliasType`, the try block in the generic path calls `get_annotation_module` &rarr; `get_annotation_class_name`, which falls back to `type(annotation).__qualname__ = "TypeAliasType"`, giving `module="typing"`, `class_name="TypeAliasType"`, and hence the wrong output.

## Expected behaviour
- If the alias is registered in the Sphinx py domain (local) or resolvable via intersphinx (external): emit :py:type: / :py:obj: cross-reference.
- Otherwise unwrap the underlying type(s).

## Other possible solutions
Use `typehints_formatter` to handle `isinstance(annotation, TypeAliasType)`

lochhh:
Closes #690 

This PR
- adds an `if isinstance(annotation, TypeAliasType):` branch in `format_annotation` to handle [typing.TypeAliasType](https://docs.python.org/3/library/typing.html#typing.TypeAliasType) introduced in python 3.12
  - the logic is similar to how `TypeAliasForwardRef` is currently handled, extending that same lookup to all intermediate prefix levels, so aliases documented anywhere in the module hierarchy resolve correctly
- adds, for external TypeAliasType (defined in a different top-level package), `_get_canonical_type_alias_name` which resolves the canonical public name
  - here we prefer public re-exports over private module paths (e.g. extpkg.ExtAlias over extpkg._priv.ExtAlias) 
  - this is used inside the `if isinstance(annotation, TypeAliasType):` branch to render a cross-reference instead of expanding to the underlying type(s)
- adds tests to cover
  - documented and undocumented `TypeAliasType` instances
  - local and external `TypeAliasType` instances

### DIFF
--- a/src/sphinx_autodoc_typehints/_annotations.py
+++ b/src/sphinx_autodoc_typehints/_annotations.py
@@ -119,6 +119,12 @@ def format_annotation(annotation: Any, config: Config, *, short_literals: bool =
             for candidate in candidates:
                 if candidate in py_domain.objects and py_domain.objects[candidate].objtype == "type":
                     return f":py:type:`{prefix}{candidate}`"
+            # Handle external type aliases
+            canonical = _get_canonical_type_alias_name(annotation)
+            current_top = module_prefix.split(".")[0] if module_prefix else ""
+            if canonical and canonical.split(".")[0] != current_top:
+                full_name = _fixup_module_name(config, canonical.rpartition(".")[0]) + "." + annotation.__name__
+                return f":py:obj:`{prefix}{full_name}`"
         return format_annotation(annotation.__value__, config, short_literals=short_literals)
 
     try:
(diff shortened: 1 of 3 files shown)


gaborbernat (approved):
LGTM

PR #705, opened 2026-06-01
process_signature: don't skip the first arg on bound instance methods

ilia-kats:
Methods bound on class instances do not have `self` as part of their signature, so skipping the first argument is wrong.


gaborbernat (approved):
LGTM

PR #716, opened 2026-06-11
✨ feat(resolver): type C data descriptors from stub files

gaborbernat:
A stub-backed C extension gets full signatures on its functions and methods, yet its properties render as bare attributes. The cause sits at the top of `process_docstring`: getset and member descriptors are not `property` instances and are not callable, so the handler returns before any stub lookup runs, even though the type is sitting right there in the `.pyi` as a `@property` return annotation or a class-level annotated assignment. I hit this documenting `turbohtml`'s C `Token` type, where `attr()` rendered with full types while the eleven properties next to it showed nothing.

The fix routes these descriptors through the same stub machinery that signatures already use. A new `get_descriptor_type_hint` finds the owning class in the stub via `__objclass__`, takes the `@property` return annotation or the matching `AnnAssign`, and evaluates it in the stub's namespace so type aliases keep their cross-referenced form. The result lands in the docstring as a `:type:` field, which the attribute directive renders the same way `:vartype:` injections do; an explicit `:type:` already present in the docstring wins, and objects without a stub stay untouched.

The `c_ext_mod` fixture grows three getset descriptors covering the `@property` return, the alias-typed union, and the class-annotation forms. Behavior for pure-Python properties is unchanged, since those keep flowing through `fget`.


gaborbernat:
Validated against real stub-shipping C extensions by running `get_descriptor_type_hint` over every getset/member descriptor they expose: cbor2, msgspec, numpy, tokenizers, pydantic_core, zstandard and rpds-py — 202 descriptors resolved, zero errors, zero false resolutions. Sample results: `CBORDecoder.tag_hook -> Optional[TagHook]` (alias kept as a cross-reference), `pydantic_core.MultiHostUrl.fragment -> str | None`, `zstandard.FrameParameters.content_size -> int`.

Every unresolved case was benign: implicit `__weakref__`/`__dict__` slots, internal classes deliberately absent from stubs, and two genuine upstream stub gaps (cbor2's `CBOREncoder.string_namespacing` and rpds' `List.first`/`List.rest` are simply not declared in their stubs). The sweep surfaced one structural edge case — numpy guards class members behind `if sys.version_info >= (3, 12):` inside `class ndarray` — so the lookup now descends guard branches, with the fixture covering a guarded `@property`.

PR #731, opened 2026-07-15
🐛 fix(overloads): keep summary above overloads

### LINKED ISSUE #730: regression in v3.8.0: `autosummary` does not work with `@overload`
ref. https://github.com/sphinx-doc/sphinx/issues/14522

```python
from typing import overload

@overload
def apply_where(...): ...

@overload
def apply_where(...): ...

def apply_where(...):
    """
    Run one of two elementwise functions depending on a condition.
    """
    ...
```

```rst
.. autosummary::
   :toctree: generated

   apply_where
```

In v3.7.0 this works fine, but since v3.8.0 this leaves the summary box blank:

Image

cc @gaborbernat

gaborbernat:
PR #625 (shipped in 3.8.0) started prepending the overload field list to a function docstring. autosummary builds its one-line summary by running `extract_summary` over the processed docstring, and that function reads the first stanza: with a `:Overloads:` field list on top it returns an empty summary rather than the real text, so autosummary boxes for overloaded functions went blank.

The fix inserts the overload block after the summary paragraph instead of at the top, so `extract_summary` still finds the summary while the overloads keep rendering in the full autodoc output.

Added a regression test that runs `extract_summary` over an injected docstring and asserts the summary survives; it fails on the old prepend behavior.

Fixes #730

### DIFF
--- a/src/sphinx_autodoc_typehints/__init__.py
+++ b/src/sphinx_autodoc_typehints/__init__.py
@@ -236,9 +236,16 @@ def _inject_overload_signatures(
         return False
     if (overloads := _resolve_overloads(obj)) is None:
         return False
-    for line in reversed(_format_overload_lines(overloads, app)):
-        lines.insert(0, line)
+    idx = _summary_end(lines)
+    lines[idx:idx] = ["", *_format_overload_lines(overloads, app)]
     return True
+
+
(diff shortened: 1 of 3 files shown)


lucascolley:
thanks for the speedy response!!

Issue #751, opened 2026-08-17
Code that is valid for a type checker but not for `exec`/interpreter.


gaborbernat:
> Ah it is here:
> 
> https://github.com/pydata/xarray/blob/abdc46498ddaf623e057c18ce35799be865bed54/xarray/core/dataarray.py#L126
> 
> ```python
> T_XarrayOther = TypeVar("T_XarrayOther", bound="DataArray" | Dataset)
> ```
> 
> Code that is valid for a type checker but not for `exec`/interpreter. 

 _Originally posted by @djhoese in [#741](https://github.com/tox-dev/sphinx-autodoc-typehints/issues/741#issuecomment-5320242657)_

djhoese:
I asked Claude if it was possible to come up with a good solution for this and it pointed out switching to a fallback that builds an AST and gets the names from that.

I then asked if it was possible to come up with a solution that backports the ForwardRef functionality in Python 3.14 and make it in a way that it could easily be removed when <3.14 is no longer supported by this tool. It didn't give me a very good answer...yet.

Let me know if you had solutions like this in mind for the future of this tool and I'll stop pursuing my own uninformed solutions (I really don't know how `sphinx-autodoc-typehints` works).

gaborbernat:
https://github.com/tox-dev/sphinx-autodoc-typehints/releases/tag/3.13.3

Issue #741, opened 2026-08-03
Version 3.13.0 now failing against TYPE_CHECKING guards


djhoese:
I have a project that has started getting warnings for import/existence issues on type annotations since sphinx-autodoc-typehints 3.13.0 started being used. We make our sphinx builds fail on any warnings it encounters but we're not sure where to go for this one. Here are some examples:

```
/home/docs/checkouts/readthedocs.org/user_builds/satpy/conda/latest/lib/python3.13/site-packages/upath/core.py:525: WARNING: Failed guarded type import in 'upath.core': ModuleNotFoundError("No module named 'pydantic'") [sphinx_autodoc_typehints.guarded_import]
/home/docs/checkouts/readthedocs.org/user_builds/satpy/conda/latest/lib/python3.13/site-packages/numpy/polynomial/_polybase.py:292: WARNING: Cannot resolve forward reference in type annotations of "satpy.readers.core.seviri.Chebyshev" (module numpy.polynomial._polybase): name '_SeriesLikeCoef_co' is not defined [sphinx_autodoc_typehints.forward_reference]
```

In the first one, `universal-pathlib` (upath) imports `pydantic` in a `if TYPE_CHECKING` guard:

https://github.com/fsspec/universal_pathlib/blob/827574045b1fecad52f7af3cfeeab3ecd9a0d0e8/upath/core.py#L62-L71

We also have warnings like:

```
/home/docs/checkouts/readthedocs.org/user_builds/satpy/conda/latest/lib/python3.13/site-packages/trollsift/parser.py:457: WARNING: Cannot resolve forward reference in type annotations of "satpy.readers.core.yaml_reader.parse" (module trollsift.parser): name 'Any' is not defined [sphinx_autodoc_typehints.forward_reference]
```

So this is our "yaml_reader.py" module in satpy importing `trollsift.parser.parse` [here](https://github.com/pytroll/satpy/blob/c16a74baee824965ed02a10136172025806d302e/satpy/readers/core/yaml_reader.py#L36). Inside the trollsift module `from typing import Any` is in a `TYPE_CHECKING` block:

https://github.com/pytroll/trollsift/blob/6ae05caf830c1d14ce6a24d6334fa9adcb3690cc/trollsift/parser.py#L14

Any idea what may have changed to cause this or how to fix it?

djhoese:
Sorry, but this doesn't seem to fix the issue. Running 3.13.2 locally and I still get the same warnings:

```
[...]

PR #757, opened 2026-08-24
🐛 fix(resolver): accept guarded code the interpreter rejects

gaborbernat:
A `TYPE_CHECKING` block only has to satisfy a type checker, so it can hold expressions the interpreter refuses. xarray writes `T_XarrayOther = TypeVar("T_XarrayOther", bound="DataArray" | Dataset)`, which raises `TypeError: unsupported operand type(s) for |: 'str' and 'type'` the moment it runs. Executing the guard to recover its names hit that error and warned `Failed guarded type import`. The annotation then reached a name nobody had bound, so a second warning followed: `Cannot resolve forward reference ... name 'T_Other' is not defined`. The rendered page was right the whole time. The warnings alone fail a build run with `-W`, which is what #751 reports.

An import that raises keeps its warning, since a dependency missing from the docs environment is worth reporting, as #741 asked for. Other statements now leave their names behind as forward references, read off the assignment targets, or off the class or function name. That matches what a type checker sees, and `format_annotation` prints such a reference as its own name, so the rendered output holds still while both warnings go away. ✨

The stand-ins go into the module namespace the guarded code writes to, through `setdefault`, so a name that already resolves keeps whatever it points at.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -131,6 +131,9 @@ lint.per-file-ignores."tests/roots/test-integration/mod_forward_ref.py" = [
   "I002",    # intentionally omits `from __future__ import annotations` to test forward references
   "PLR6301", # method must be instance method to test class forward refs
 ]
+lint.per-file-ignores."tests/roots/test-unexecutable-guard/demo_unexecutable_guard.py" = [
+  "UP047", # the legacy TypeVar spelling is what the guard cannot execute
+]
 lint.per-file-ignores."tests/roots/test-pyi-stubs/stub_mod.py" = [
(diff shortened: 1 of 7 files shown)


djhoese:
Wow, awesome, thanks for the fixes. One question:

> An import that raises keeps its warning, since a dependency missing from the docs environment is worth reporting

Does this mean that if a type checking guard has a `try/except ImportError` that it will still warn on the missing import since I think this code is still executing one line at a time? Installing these missing dependencies is an easy enough workaround, but I wanted to check.

Issue #764, opened 2026-09-01
`_format_node` crashes when referencing a type alias in a different file


flying-sheep:
This is order-dependent, so if `pkg._types.g` in the repro is encountered first, it doesn’t occur

https://github.com/tox-dev/sphinx-autodoc-typehints/blob/60d74b3765c2b730e8ca34f19a96e8a1971e46aa/src/sphinx_autodoc_typehints/_annotations.py#L174

## repro files

`conf.py`
```py
import sys
from pathlib import Path

sys.path.insert(0, str(Path(__file__).parent))

project = "repro"
extensions = ["sphinx.ext.autodoc", "sphinx_autodoc_typehints"]
```

`index.rst`
```rst
Repro
=====

.. autofunction:: pkg.api.f
```

`pkg/__init__.py`: empty

`pkg/api.py`
```py
api.py 
"""The documented module: it uses the alias, but does not define it."""

from __future__ import annotations

from typing import TYPE_CHECKING

if TYPE_CHECKING:
    from ._types import Alias


def f(x: Alias | None = None) -> None:
    """Do nothing.

    :param x: an argument.
    """
```

`pkg/_types.py`
```py
"""Defines an (undocumented) PEP 695 alias whose value needs a type-guarded import."""

from __future__ import annotations

from typing import TYPE_CHECKING

if TYPE_CHECKING:
    from collections.abc import Sequence

type Alias = int | Sequence[str]


def g() -> None:
    """Owned by the module that defines the alias."""
```

## repro

`sphinx-build -M html . _build`

```pytb
File ".../sphinx_autodoc_typehints/_annotations.py", line 174, in _format_node
    return (yield annotation.__value__)
File ".../pkg/_types.py", line 10, in Alias
    type Alias = int | Sequence[str]
NameError: name 'Sequence' is not defined
```

PR #765, opened 2026-09-01
🐛 fix(annotations): expand aliases from unread modules

gaborbernat:
Building docs for a function annotated with a PEP 695 alias defined in another module crashed with `NameError: name 'Sequence' is not defined`, as reported in https://github.com/tox-dev/sphinx-autodoc-typehints/issues/764. An alias evaluates its value lazily in the namespace of the module that defines it, so `type Alias = int | Sequence[str]` needs that module's `if TYPE_CHECKING` block to have run. We execute those guarded imports only for modules autodoc reaches directly, which is why documenting anything from the defining module first made the error disappear and print the expanded value instead.

`_format_node` now runs the defining module's guarded imports before expanding an alias whose value does not evaluate, so `pkg.api.f` renders `int | Sequence[str] | None` whatever order Sphinx reads the two modules in. 🔁 An alias whose value stays unevaluatable falls back to a `:py:type:` cross-reference to its own name, the same fallback a wrong-arity generic subscript already takes.

The resolver reaches the formatter through a callable published on the config, alongside the build environment and module prefix it already publishes there. That keeps `_annotations` importing nothing from the rest of the package, and the work stays lazy: nothing extra runs until a value fails to evaluate.

### DIFF
--- a/src/sphinx_autodoc_typehints/__init__.py
+++ b/src/sphinx_autodoc_typehints/__init__.py
@@ -6,6 +6,7 @@ import inspect
 import re
 import types
 from contextlib import contextmanager
+from functools import partial
 from typing import TYPE_CHECKING, Any, TypeVar
 
 from docutils import nodes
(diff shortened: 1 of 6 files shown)


flying-sheep:
thank you!

PR #766, opened 2026-09-04
🐛 fix(annotations): defer alias expansion

flying-sheep:
Get rid of the order dependency of the alias expansion.

Now a type alias gets expanded when it’s not documented and not expanded when it’s documented.

After this PR (or as part of this PR), the code base could be simplified a bit by always deferring and resolving everything in the end, instead of opportunistically resolving things that can be resolved and deferring the rest.

### DIFF
--- a/src/sphinx_autodoc_typehints/_parser.py
+++ b/src/sphinx_autodoc_typehints/_parser.py
@@ -2,7 +2,8 @@
 
 from __future__ import annotations
 
-from typing import TYPE_CHECKING
+import re
+from typing import TYPE_CHECKING, Final
 
 from docutils.utils import new_document
 from sphinx.parsers import RSTParser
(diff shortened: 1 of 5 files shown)


src/sphinx_autodoc_typehints/_parser.py
```diff
@@ -28,3 +37,11 @@ def parse(inputstr: str, settings: Values | optparse.Values) -> nodes.document:
     # losing the external+ roles the read phase resolves (#753)
     _RstSnippetParser().parse(inputstr, doc)
     return doc
+
+
+def unescape(escaped: str) -> str:
+    escaped = escaped.replace("\x00", "")
+    return _UNESCAPE_RE.sub(r"\1", escaped)
```
flying-sheep, line 44: moved this over here to prevent circular imports, and I think it fits here pretty well anyway.

tests/test_pep695.py
```diff
@@ -571,7 +571,8 @@ def some_func(some_param: RecType) -> None:
     assert "reference target not found" not in warning.getvalue()
 
     result = normalize_sphinx_text((Path(app.srcdir) / "_build/text/index.txt").read_text())
-    assert '"int" | "list"["RecType"]' in result
+    # documented further down the same document, which only the resolve phase can know
+    assert '**some_param** ("RecType")' in result
```
flying-sheep, line 575: this test changed because we’re now better at not expanding things that are documented.

gaborbernat (approved):

